PR #93, opened 2017-04-18
Fix getting started url

plastiv:
Previous url is resulted in 404.


plastiv:
@facebook-github-bot signed

passy:
@plastiv The first public PR! 🎉  Thank you.

PR #103, opened 2017-04-20
Fix search bar style

CodinCat:
The search input doesn't align correctly.

before

after


passy:
Not sure why that was added to begin with, but it does definitely look better without. Thanks!

PR #107, opened 2017-04-20
Update isSticky() with required parameter

pavlospt:
`ComponentInfo` method `isSticky()` accepts a boolean parameter.


passy:
Great catch! Thanks!

PR #110, opened 2017-04-21
Fix typo in documentation (uses.md)

jimfb:
Proper english grammar dictates the use of the article "a" instead of "an", despite the fact that the following word begins with a vowel.  Yes, the rules of english grammar are confusing.

For reference, look at "Use 5" at: http://www.englishpage.com/articles/a-vs-an.htm


jimfb:
Silly @facebook-github-bot.  Apparently my CLA record got confused when I left facebook.  I just sent an email to cla@fb.

passy:
This was fixed now. Thanks again for pointing this out. :)

PR #117, opened 2017-04-22
Use public framework name (Litho) instead of internal framework name (Components)

jimfb:
I'm assuming we're intending to use Litho as the public name for this library, and it was internally referred to as Components, since it appears to be a parallel of ComponentKit/Components for iOS.  Under this assumption, the phrase "Component framework" should probably be renamed "Litho framework".


grandstaish:
@jimfb There's also an error message on line 69 of AbstractComponentsProcessor.java which refers to "the components team." Might as well update that too as a part of this change.

passy:
Fixed, too.

PR #121, opened 2017-04-22
Typo: Should be `@OnUpdateState` instead of `@OnUpdateMethod`

jimfb:
Presumably this was supposed to be `@OnUpdateState`.

### DIFF
--- a/docs/_docs/state.md
+++ b/docs/_docs/state.md
@@ -80,7 +80,7 @@ You can define how a component's state or states should updated by declaring met
 
 You can have as many `@OnUpdateState` methods as you need, according to what states you want to update or what parameters your states depend on.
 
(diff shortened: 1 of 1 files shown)


docs/_docs/state.md
```diff
@@ -80,7 +80,7 @@ You can define how a component's state or states should updated by declaring met
 
 You can have as many `@OnUpdateState` methods as you need, according to what states you want to update or what parameters your states depend on.
 
-Each call to an `@OnUpdateState` method will trigger a new layout calculation for its ComponentTree. For better performance, if there are situations that can trigger an update for multiple states, you should define an `@OnUpdateMethod` that updates the value for all those states. Bundling them in the same update call reduces the number of new layout calculations and improves performance.
+Each call to an `@OnUpdateState` method will trigger a new layout calculation for its ComponentTree. For better performance, if there are situations that can trigger an update for multiple states, you should define an `@OnUpdateState` that updates the value for all those states. Bundling them in the same update call reduces the number of new layout calculations and improves performance.
```
mihaelao: good catch. could you please change to "you should define an `@OnUpdateState` method"?

passy:
Fixed in master.

PR #124, opened 2017-04-23
Add GlideImageSpec and modify sample to support it

pavlospt:
Hello,

After reading this article http://www.jayrambhia.com/blog/android-litho-gifs we (me & @charbgr) took the initiative to build a Glide module for Litho that can support Image loading using Glide. Since it seems that people use it a lot it might be a good addition to Litho.

### Things missing
* BUCK file (since I am not familiar with it)

### Other thoughts
With pretty much the same work, we can add a litho-picasso module as well.

@lucasr This is what I was telling you about on Twitter :) 

Let us know what you think and looking forward to your constructive feedback!

P.S.: It seems like we have broken some code styles, so if there is an available code style file for Android Studio that you use, let us know so that we can import it and correct style issues!


passy:
@pavlospt Woah, this is amazing! 🎉  After reading the blog post I had the same thought, that it would definitely be useful to have this as stand-alone component.

We've realized that we have no communication in the repository right now indicating what should go in and what should go elsewhere. Something we'll definitely address shortly.

A Glide component is one of those things we believe would serve best as stand-alone library in its own repository. Frankly, we'd like to have `litho-fresco` as separate repo but our current tooling doesn't support this in a convenient way.

How would you feel about turning this into a separate repo? That way we would also avoid the question which image library to use for the sample app and use both. :)

pavlospt:
@passy Sure that would be great :) I can make a separate repository :)

passy:
@pavlospt Awesome! Let me close this here, but please ping me or open a new issue when it's ready. We can create a page in the docs with third-party integrations and link to it from there. :)

pavlospt:
@passy Will ping you here, when it's done :)

pavlospt:
@passy It's ready :) https://github.com/pavlospt/litho-glide Have a look if you want :)

PR #127, opened 2017-04-24
Remove redundant dependencies.

runningcode:
The support-v4 library was modularized into distinct components like
support-compat, support-core-ui, support-fragment. We should not depend
on the modules as well as the unmodularized version.

I got errors when running a buck build unrelated to this change so ill need to verify that this compiles with buck. `./gradlew assemble` succeeds.

### DIFF
--- a/.buckconfig
+++ b/.buckconfig
@@ -9,4 +9,4 @@
 
 [android]
   target = android-25
-  build_tools_version = 23.0.2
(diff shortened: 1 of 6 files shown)


passy:
@runningcode Thanks! This was to keep things in line with the internal build, but this should be backwards-compatible. Let me try running this on the internal CI. :)

Issue #79, opened 2017-04-12
Use bit flag instead of booleans in TestDrawableComponent


emilsjolander:
We have a lot of booleans to configure the test components like `callsShouldUpdateOnMount` or `implementsAccessibility`. Instead of keeping the configuration flags as booleans we could have them as bit flags.

### Code pointers
https://github.com/facebook/litho/blob/master/litho-testing/src/main/java/com/facebook/litho/testing/TestDrawableComponent.java

Point of contact: @mihaelao

anu0012:
Hey @emilsjolander, I want to take up this issue. Can you guide me? Do we need to use bit flags just like it is used in InternalNode.java file?

https://github.com/facebook/litho/blob/master/litho-core/src/main/java/com/facebook/litho/InternalNode.java

emilsjolander:
Awesome! Yes I believe so. Will delegate specifics to @mihaelao who is the POC for this issue 👍  She should be able to provide more specific guidance.

anu0012:
Hey @mihaelao, I have some doubt regarding this issue. Do we also need to change the return types of methods like callsShouldUpdateOnMount() etc?

mihaelao:
Hi @anu0012, thanks for taking this up! You're right, the idea is to use bit flags instead of booleans, same as for InternalNode. The API shouldn't change though, those methods should still return booleans, but you can perform bitwise operations on the flags instead of returning the booleans directly.

muraziz:
Resolved with https://github.com/facebook/litho/issues/138

PR #138, opened 2017-04-26
Replaced boolean flags with EnumSet

sina-al:
Instance boolean flags were replaced by a single `EnumSet<Property>`.  The `Property` enum was nested in this class. NB: the instance field mProperties is mutable whereas its boolean predecessors were final.

### DIFF
--- a/litho-testing/src/main/java/com/facebook/litho/testing/TestDrawableComponent.java
+++ b/litho-testing/src/main/java/com/facebook/litho/testing/TestDrawableComponent.java
@@ -10,7 +10,9 @@
 package com.facebook.litho.testing;
 
 import java.util.ArrayList;
+import java.util.EnumSet;
 import java.util.List;
+import java.util.Set;
 
 import android.graphics.Color;
 import android.graphics.drawable.ColorDrawable;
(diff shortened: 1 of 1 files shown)


muraziz:
Hi @sina-al, thanks for your efforts! We would like to use bit flags instead of enums, can you follow-up the thread at https://github.com/facebook/litho/issues/79?

sina-al:
Hello, sorry. I have attempted to implement bit flags instead.

litho-testing/src/main/java/com/facebook/litho/testing/TestDrawableComponent.java
```diff
@@ -224,17 +197,23 @@ public static Builder create(
...
-            callsShouldUpdateOnMount,
-            isPureRender,
-            canMeasure,
-            implementsAccessibility,
-            usesDisplayList,
-            isMountSizeDependent));
+
+    byte properties = 0;
+    
+    if (callsShouldUpdateOnMount)
```
muraziz: Just to follow common coding style can you add brackets? We usually don't use bracketless if statements.

litho-testing/src/main/java/com/facebook/litho/testing/TestDrawableComponent.java
```diff
@@ -30,59 +30,32 @@
...
-      boolean usesDisplayList,
-      boolean isMountSizeDependent) {
+  private static final byte CALLS_SHOULD_UPDATE_ON_MOUNT = 1 << 0;
+  private static final byte IS_PURE_RENDER = 1 << 1;
+  private static final byte CAN_MEASURE = 1 << 2;
+  private static final byte USES_DISPLAY_LIST = 1 << 3;
+  private static final byte IMPLEMENTS_ACCESSIBILITY = 1 << 4;
+  private static final byte IS_MOUNT_SIZE_DEPENDENT = 1 << 5;
+
+  private byte mProperties;
```
muraziz: `byte` allows us to keep only 8 flags, but we already have 6 values, so let's use `long` to have enough space for future flags. This will involve changing the flag themselves above as well.

sina-al:
Hello @muraziz 

Those suggestions make sense. Thanks for explaining them, and thanks for the patience as this is my first pull request. 

I also have a question about spacing/indenting:

I managed to change the settings on my IDE to change the indenting to two spaces (as opposed to four) but i still found some lines that I had to adjust manually. How would one fully set the code style so it matches the one on this repo? (If you look between the commits on my fork, you will be able to see the specific type of indenting I'm referencing. Apologies, I am not aware of the correct terminology).

Kind Regards,
Sina

muraziz:
Hi @sina-al,

I don't think we have an IDE-level guideline for our codebase, but we have the one more tailored to our framework which is described in the docs: http://fblitho.com/docs/best-practices#coding-style. We use unix line break and 100 char limit for single line.

With your last commit it looks fine, I'll land them. Thanks for your contribution!

passy:
@sina-al Great patch! Thanks for doing that.

PR #165, opened 2017-05-02
Fixes #151

### LINKED ISSUE #151: ClassCastException during annotation processing, triggered by package-level annotations

At build time annotation processing, the Litho AbstractComponentsProcessor runs into a ClassCastException. Here is a partial stack trace:
```
java.lang.ClassCastException: com.sun.tools.javac.code.Symbol$PackageSymbol cannot be cast to javax.lang.model.element.TypeElement
    at com.facebook.litho.specmodels.processor.AbstractComponentsProcessor.process(AbstractComponentsProcessor.java:48)
```
Here is the relevant line of code:
https://github.com/facebook/litho/blob/cbbd4e26129e45ab6e43bb70473a2f8a65d600c3/litho-processor/src/main/java/com/facebook/litho/specmodels/processor/AbstractComponentsProcessor.java#L48

This is triggered by the package-level annotation `@Gson.TypeAdapters` we are using from the Immutables library:
https://immutables.github.io/json.html#generating-type-adapters

It looks like AbstractComponentsProcessor makes an assumption that the annotation is for a class or interface, but the case we're using the package-level annotation, it's a fragile assumption.

Repro steps:
1. Bring in the immutables gson library, i.e. `provided 'org.immutables:gson:2.4.6'`
2. Create a Litho spec. The `@LayoutSpec` annotation is sufficient.
3. Add the `@Gson.TypeAdapters` annotation in a new package-info.java file, e.g.:
```java
@Gson.TypeAdapters
package com.your.package.name;

import org.immutables.gson.Gson;
```
4. Attempt to build with the litho-processor, i.e. with `annotationProcessor 'com.facebook.litho:litho-processor:0.2.0'`

## Expected Behavior
I expect `AbstractComponentsProcessor` to not process the annotation if the `element` is not an instance of `TypeElement`, allowing apps which use package-level annotations to succeed.

## Link to Code
(I can work on a minimal app if the above repro steps are insufficient.)

pavlospt:
Fixes #151 by getting only the `LayoutSpec` and `MountSpec` annotation classes for processing.

### DIFF
--- a/litho-processor/src/main/java/com/facebook/litho/specmodels/processor/AbstractComponentsProcessor.java
+++ b/litho-processor/src/main/java/com/facebook/litho/specmodels/processor/AbstractComponentsProcessor.java
@@ -42,7 +42,16 @@ public abstract class AbstractComponentsProcessor extends AbstractProcessor {
       return false;
     }
 
-    for (Element element : roundEnv.getRootElements()) {
+    Set<Element> layoutSpecElements =
+        (Set<Element>) roundEnv.getElementsAnnotatedWith(LayoutSpec.class);
+    Set<Element> mountSpecElements =
+        (Set<Element>) roundEnv.getElementsAnnotatedWith(MountSpec.class);
+
+    Set<Element> allSpecElements = new HashSet<>();
+    allSpecElements.addAll(layoutSpecElements);
+    allSpecElements.addAll(mountSpecElements);
+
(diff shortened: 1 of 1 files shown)


IanChilds:
Looks great. Thanks!

passy:
Thank you, @pavlospt! 🎉

PR #167, opened 2017-05-03
Add missing static modifier on updateCheckBox

### LINKED ISSUE #166: Maybe a bug in http://fblitho.com/docs/state#calling-state-updates

In the piece of code in http://fblitho.com/docs/state#calling-state-updates,
 @OnUpdateState
  void updateCheckbox(StateValue<Boolean> isChecked) {
    isChecked.set(!isChecked.get());
  }
is void, but I thought other method in spec is static, so the function can not work with others.
Be more clear, I wrote a sample project use the code, it fails when build, and tells me the method should be static.
So maybe it's a mistake.

pavlospt:
Fixes #166

### DIFF
--- a/docs/_docs/state.md
+++ b/docs/_docs/state.md
@@ -144,7 +144,7 @@ public class CheckboxSpec {
   }
 
   @OnUpdateState
-  void updateCheckbox(StateValue<Boolean> isChecked) {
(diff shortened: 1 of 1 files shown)


passy:
Good catch! Thanks!

Issue #81, opened 2017-04-12
Support prop defaults being resource identifiers


emilsjolander:
Currently there is no way to specify the default value of a `ResType` prop using resource ids.

Say you declare a prop like the following
```java
@Prop(optional = true, resType = ResType.DIMEN_SIZE) int textSize
```

You would like to be able to declare its default value as being a dimension resource.
```java
@PropDefault protected static final int textSize = R.dimen.text_size_medium;
```

However that currently would not work as it would parse the resource id as an actual size value and not grab the size associated with the id.

This would most likely be solved by adding `resType` to the `@PropDefault` annotation.
```java
@PropDefault(resType = ResType.DIMEN_SIZE) protected static final int textSize = R.dimen.text_size_medium;
```

### Code pointers
https://github.com/facebook/litho/blob/master/litho-annotations/src/main/java/com/facebook/litho/annotations/Prop.java

https://github.com/facebook/litho/blob/master/litho-annotations/src/main/java/com/facebook/litho/annotations/PropDefault.java

https://github.com/facebook/litho/blob/master/litho-processor/src/main/java/com/facebook/litho/specmodels/generator/ComponentImplGenerator.java#L158

Point of contact: @emilsjolander

pavlospt:
I have started working on this and my progress so far is: 
* I have added the ability to add a ResType on a PropDefault annotation.
* I extract it successfully and add it on the PropDefaultModel
* Parsing it correctly and initializing the Prop field with a proper initializer (only using the `resolveDimenSizeRes()` for now)

My thoughts on the above: 
* In order to use the resource resolve method, I tweaked the Component class to extend ResourceResolver and code generation works properly now. Although what this does is crash on Runtime because it does not have a valid ComponentContext to call `super.init(c, c.getResourceCache())`. 

I do not know if I am on the right path yet, but it seems to me that in order to support this we need to either extend ResourceResolver from Component and provide it with a valid ComponentContext or just initialize the PropDefault marked variable on the Builder where we have a valid ComponentContext.

Let me know what you think.

/cc: @emilsjolander @passy 

P.S.: Sorry for pinging here as well, might seem like spamming, but I want to save the time for anyone who would like to start working on this one and we end up having duplicate work.

emilsjolander:
Could you publish your current progress as a pull request? This way we could make better suggestions based on the code. From the info you provided I think both alternatives would work. I see no reason not to pass the context down to the component.

pavlospt:
[...]

PR #174, opened 2017-05-04
Support ResType on PropDefault annotation

pavlospt:
Initial work on adding support for `ResType` on `@PropDefault` ( #81 )

I went with the solution of having the `Builder` initializing the Props that have a default value marked as ResType, instead of passing down the `ComponentContext` which I tried to implement but it was introducing a quite huge change.

In order to reduce the review scope I have only added generation for `resolveDimenSizeRes()` and will add the rest before closing the PR, in order to have a naming and feature compliance agreement after the initial review. Also, I believe that after resolving the resource on a marked `@PropDefault` we should generate resource helper methods like we do for `@Prop`.

Let me know what you think! 

/cc: @emilsjolander 

P.S.: There is a big diff on the `ComponentImplGenerator` due to different Code Style, so If you could share the one you are using so that we could be aligned and reduce the diffs, would be great.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,4 +18,9 @@ docs/.bundle/
 # Android Studio/IntelliJ
 .idea
 *.iml
+.gradle/
+lib/fb/
+lib/yoga/
+lib/yogajni/
+*/**.iml
 local.properties
--- a/.gitignore
+++ b/.gitignore
@@ -18,9 +18,4 @@ docs/.bundle/
 # Android Studio/IntelliJ
 .idea
 *.iml
-.gradle/
-lib/fb/
-lib/yoga/
-lib/yogajni/
-*/**.iml
 local.properties
--- a/litho-processor/src/main/java/com/facebook/litho/specmodels/generator/BuilderGenerator.java
+++ b/litho-processor/src/main/java/com/facebook/litho/specmodels/generator/BuilderGenerator.java
@@ -126,11 +126,11 @@ public class BuilderGenerator {
         .addParameter(specModel.getContextClass(), CONTEXT_PARAM_NAME);
 
     final ImmutableList<PropDefaultModel> propDefaults = specModel.getPropDefaults();
-    boolean hasResTypePropDefault = false;
+    boolean isResResolvable = false;
 
     for (PropDefaultModel propDefault : propDefaults) {
-      if (propDefault.hasResType()) {
-        hasResTypePropDefault = true;
+      if (propDefault.isResResolvable()) {
+        isResResolvable = true;
         break;
       }
     }
@@ -151,7 +151,7 @@ public class BuilderGenerator {
         .addStatement("$L = $L", implMemberInstanceName, implParamName)
         .addStatement("$L = $L", CONTEXT_MEMBER_NAME, CONTEXT_PARAM_NAME);
 
-    if (hasResTypePropDefault) {
+    if (isResResolvable) {
       initMethodSpec.addStatement("initPropDefaults()");
     }
 
@@ -218,15 +218,15 @@ public class BuilderGenerator {
 
     MethodSpec.Builder initResTypePropDefaultsSpec = null;
 
-    if (hasResTypePropDefault) {
+    if (isResResolvable) {
       initResTypePropDefaultsSpec = MethodSpec.methodBuilder("initPropDefaults");
 
       for (PropDefaultModel propDefault : propDefaults) {
-        if (!propDefault.hasResType()) continue;
+        if (!propDefault.isResResolvable()) continue;
 
         initResTypePropDefaultsSpec.addStatement("this.$L.$L = $L",
-            getImplMemberInstanceName(specModel), propDefault.mName,
-            generateInitializer(propDefault, specModel));
+            getImplMemberInstanceName(specModel), propDefault.getName(),
+            generatePropsDefaultInitializers(specModel, propDefault));
       }
     }
 
@@ -436,6 +436,40 @@ public class BuilderGenerator {
     }
 
     return dataHolder.build();
+  }
+
+  static String generatePropsDefaultInitializers(
+      SpecModel specModel,
+      PropDefaultModel propDefault) {
+
+    switch (propDefault.getResType()) {
+      case STRING:
+        return generatePropDefaultResInitializer("resolveStringRes", propDefault, specModel);
+      case STRING_ARRAY:
+        return generatePropDefaultResInitializer("resolveStringArrayRes", propDefault, specModel);
+      case INT:
+        return generatePropDefaultResInitializer("resolveIntRes", propDefault, specModel);
+      case INT_ARRAY:
+        return generatePropDefaultResInitializer("resolveIntArrayRes", propDefault, specModel);
+      case BOOL:
+        return generatePropDefaultResInitializer("resolveBoolRes", propDefault, specModel);
+      case COLOR:
+        return generatePropDefaultResInitializer("resolveColorRes", propDefault, specModel);
+      case DIMEN_SIZE:
+        return generatePropDefaultResInitializer("resolveDimenSizeRes", propDefault, specModel);
+      case DIMEN_TEXT:
+        return generatePropDefaultResInitializer("resolveDimenSizeRes", propDefault, specModel);
+      case DIMEN_OFFSET:
+        return generatePropDefaultResInitializer("resolveDimenOffsetRes", propDefault, specModel);
+      case FLOAT:
+        return generatePropDefaultResInitializer("resolveFloatRes", propDefault, specModel);
+      case DRAWABLE_REFERENCE:
+        return generatePropDefaultResInitializer("resolveDrawableReferenceRes", propDefault, specModel);
+      case DRAWABLE:
+        return generatePropDefaultResInitializer("resolveDrawableRes", propDefault, specModel);
+    }
+
+    return "";
   }
 
   static TypeName getRawType(TypeName type) {
@@ -814,19 +848,20 @@ public class BuilderGenerator {
         .build();
   }
 
-  private static String generateInitializer(
+  private static String generatePropDefaultResInitializer(
+      String resourceResolveMethodName,
       PropDefaultModel propDefaultModel,
       SpecModel specModel) {
-    final String defaultInitializer = "%s.%s";
     StringBuilder builtInitializer = new StringBuilder();
 
-    if (propDefaultModel.hasResType()) {
-      if (propDefaultModel.mResType == ResType.DIMEN_SIZE) {
-        return String.format(builtInitializer.append("resolveDimenSizeRes(")
-            .append(defaultInitializer)
-            .append(")")
-            .toString(), specModel.getSpecName(), propDefaultModel.mName);
-      }
+    if (propDefaultModel.isResResolvable()) {
+      return String.format(
+          builtInitializer
+              .append(resourceResolveMethodName)
+              .append("(")
+              .append(propDefaultModel.getResId())
+              .append(")")
+              .toString(), specModel.getSpecName(), propDefaultModel.getName());
     }
 
     return builtInitializer.toString();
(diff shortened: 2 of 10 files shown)


emilsjolander:
@pavlospt Thanks! I'll have a look today or tomorrow and give more detailed feedback. Before that could you please revert the style changes you made? We have our preferred formatting style and would rather not change it, also makes the PR harder to review.

pavlospt:
@emilsjolander Yeap that's what I wrote on the ps, could you share your formatting style so that I can revert it?  Because right now I don't see how I could that. Let me know if you have any ideas!

pavlospt:
Code style issues are fixed!

emilsjolander:
We don't have a formal formatting style really. We should probably work on one though 👍  Just look at how the rest of the code in the files are formatted and try to follow the same style.

sample/src/main/java/com/facebook/samples/litho/lithography/FeedItemCardSpec.java
```diff
@@ -8,8 +8,10 @@
 
 package com.facebook.samples.litho.lithography;
 
+import android.graphics.drawable.Drawable;
```
emilsjolander: Seems unused

sample/src/main/java/com/facebook/samples/litho/lithography/FeedItemCardSpec.java
```diff
@@ -8,8 +8,10 @@
...
 import com.facebook.litho.Column;
 
+import com.facebook.litho.annotations.ResType;
```
emilsjolander: seems unused

sample/src/main/java/com/facebook/samples/litho/playground/PlaygroundComponentSpec.java
```diff
@@ -21,13 +23,16 @@
 
 import com.facebook.litho.annotations.LayoutSpec;
 import com.facebook.litho.annotations.OnCreateLayout;
-import com.facebook.litho.widget.Text;
 
 @LayoutSpec
 public class PlaygroundComponentSpec {
 
+  @PropDefault(resType = ResType.DIMEN_SIZE)
```
emilsjolander: Could you revert these changes? Good to see example used but we don't want to add a bunch of stuff to the playground.

sample/src/main/res/values/dimens.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
```
emilsjolander: Remove this file.

litho-processor/src/main/java/com/facebook/litho/specmodels/generator/BuilderGenerator.java
```diff
@@ -733,4 +768,21 @@ private static MethodSpec generateReleaseMethod(SpecModel specModel) {
         .addStatement("$L.release(this)", BUILDER_POOL_FIELD)
         .build();
   }
+
+  private static String generateInitializer(PropDefaultModel propDefaultModel,
```
emilsjolander: If a line exceeds 100 we put all the parameters on the next line, including the first one 👍

litho-processor/src/main/java/com/facebook/litho/specmodels/generator/BuilderGenerator.java
```diff
@@ -733,4 +768,21 @@ private static MethodSpec generateReleaseMethod(SpecModel specModel) {
...
+      SpecModel specModel) {
+    final String defaultInitializer = "%s.%s";
+    StringBuilder builtInitializer = new StringBuilder();
+
+    if (propDefaultModel.hasResType()) {
+      if (propDefaultModel.mResType == ResType.DIMEN_SIZE) {
```
emilsjolander: We should handle all res types. See how this is done for the builder methods.

pavlospt:
@emilsjolander thank you for the review, will try to fix them and finalize it.

litho-processor/src/main/java/com/facebook/litho/specmodels/generator/BuilderGenerator.java
pavlospt: @emilsjolander Do you mean 100 chars? This line seems to be 78 chars including the leading space.

pavlospt:
@emilsjolander Fixed most of the review changes, one thing that is not clear though: in order to generate the initializers, should I create new variables? 

What I mean is: 

* If someone marks a `@PropDefault` with a `ResType` then it would most probably require it to be an `int` in order to accept the `R` reference, which means that we would need to generate a variable for the resolved type of the `ResType`.

pavlospt:
I had a discussion with @charbgr for possible approaches on the above issue: 

The outcome was that we could add another key on `@PropDefault` called `resId` possibly and there the user would declare the `resourceId` that we would resolve. 

That means that the user gets to declare the same type of variable as the `@Prop` annotated and we would not need to create other variables as mediators.

emilsjolander:
Looks pretty good! thanks 👍  Could you please add tests for the added features?

pavlospt:
@emilsjolander Sure thing, is there a specific directory for the Annotation Processor tests ?

passy:
Oh shoot, we're still working on getting the annotation processor tests open-sourced. There were some issues with Gradle and @IanChilds is working on fixing those.

pavlospt:
@passy Hmmm, is there any way around this then ?

IanChilds:
We should have them available today or tomorrow :)

pavlospt:
@IanChilds Great!

nscoding:
Any updates here? would be great to see this merged.

pavlospt:
@nscoding Did not find some free time to write the tests yet..Hopefully will be able to do so, in the following week!

pavlospt:
@IanChilds can you please review the last commit? I had to use `12345` as a constant for the `BuildGeneratorTest` additions due to limitation annotation fields that need to be a constant expression. Inside a ComponentSpec it works like a charm because the R field is resolved into a constant ID in advance.

.gitignore
```diff
@@ -18,4 +18,9 @@ docs/.bundle/
 # Android Studio/IntelliJ
 .idea
 *.iml
+.gradle/
+lib/fb/
```
passy: We need those. :)
Can you revert that?

.gitignore
```diff
@@ -18,4 +18,9 @@ docs/.bundle/
...
+lib/yoga/
+lib/yogajni/
+*/**.iml
```
passy: This should already be covered by the one above.

litho-processor/src/main/java/com/facebook/litho/specmodels/generator/BuilderGenerator.java
```diff
@@ -199,8 +216,26 @@ static TypeSpecDataHolder generateBuilder(SpecModel specModel) {
...
 
+    MethodSpec.Builder initResTypePropDefaultsSpec = null;
+
+    if (isResResolvable) {
+      initResTypePropDefaultsSpec = MethodSpec.methodBuilder("initPropDefaults");
+
+      for (PropDefaultModel propDefault : propDefaults) {
+        if (!propDefault.isResResolvable()) continue;
+
+        initResTypePropDefaultsSpec.addStatement("this.$L.$L = $L",
```
IanChilds: one param per line please :)

litho-processor/src/main/java/com/facebook/litho/specmodels/generator/BuilderGenerator.java
```diff
@@ -199,8 +216,26 @@ static TypeSpecDataHolder generateBuilder(SpecModel specModel) {
...
+            getImplMemberInstanceName(specModel), propDefault.getName(),
+            generatePropsDefaultInitializers(specModel, propDefault));
+      }
+    }
+
     propsBuilderClassBuilder.addMethod(initMethodSpec.build());
 
+    if (initResTypePropDefaultsSpec != null) {
```
IanChilds: why don't you just do everything at this point and then you don't have to do the null checks?

litho-processor/src/main/java/com/facebook/litho/specmodels/model/PropDefaultModel.java
```diff
@@ -53,6 +96,8 @@ public int hashCode() {
     int result = mType.hashCode();
     result = 17 * result + mName.hashCode();
     result = 31 * result + mModifiers.hashCode();
+    result = 31 * result + mResType.hashCode();
```
IanChilds: use a different number here and on the line below

litho-processor/src/main/java/com/facebook/litho/specmodels/model/PropModel.java
```diff
@@ -97,6 +97,17 @@ public boolean hasDefault(ImmutableList<PropDefaultModel> propDefaults) {
     return false;
   }
 
+  public PropDefaultModel getDefault(ImmutableList<PropDefaultModel> propDefaults) {
```
IanChilds: what is this for?

litho-processor/src/main/java/com/facebook/litho/specmodels/generator/BuilderGenerator.java
pavlospt: @IanChilds can you explain this a little bit more? :)

litho-processor/src/main/java/com/facebook/litho/specmodels/model/PropModel.java
pavlospt: @IanChilds unused, I removed it. Thanks!

lucasr:
I'd update the docs for PropDefaults (https://fblitho.com/docs/props#prop-defaults) to cover this new feature :-)

pavlospt:
@lucasr I can do this as well and include it in this PR :)

litho-processor/src/main/java/com/facebook/litho/specmodels/generator/BuilderGenerator.java
pavlospt: Done!

litho-processor/src/main/java/com/facebook/litho/specmodels/generator/BuilderGenerator.java
ahmedre: i think what he's saying is if you move this up to the bottom of the if on line 221, you can drop the null check.

litho-processor/src/main/java/com/facebook/litho/specmodels/generator/BuilderGenerator.java
```diff
@@ -199,8 +216,26 @@ static TypeSpecDataHolder generateBuilder(SpecModel specModel) {
       initMethodSpec.addStatement("mRequired.clear()");
     }
 
+    MethodSpec.Builder initResTypePropDefaultsSpec = null;
```
ahmedre: move into the if statement below

sample/src/main/java/com/facebook/samples/litho/playground/PlaygroundComponentSpec.java
```diff
@@ -10,9 +10,7 @@
 
 import com.facebook.litho.Column;
 
-import com.facebook.yoga.YogaAlign;
-
-import com.facebook.yoga.YogaFlexDirection;
+import com.facebook.litho.widget.Text;
```
ahmedre: nitpick: can you undo this change (it's nicer when we keep them alphabetized).

litho-processor/src/main/java/com/facebook/litho/specmodels/model/PropDefaultModel.java
```diff
@@ -53,6 +96,8 @@ public int hashCode() {
     int result = mType.hashCode();
     result = 17 * result + mName.hashCode();
     result = 31 * result + mModifiers.hashCode();
+    result = 43 * result + mResType.hashCode();
+    result = 65 * result + mResId;
```
ahmedre: nitpick - for consistency, can you replace 65 with 47 or 53 or 59 or 61 to keep it as prime numbers?

pavlospt:
@ahmedre Updated to reflect review requests. Can you check again :) ?

ahmedre:
@pavlospt may you also update [props.md](https://github.com/facebook/litho/blob/master/docs/_docs/props.md) with these changes? btw, thanks for doing this!

lgtm otherwise, but will leave the remaining review checks to the Litho core team

pavlospt:
@ahmedre Done :)

PR #176, opened 2017-05-06
Add realpath apt package

pavlospt:


passy:
Hah, thanks, but I'll just just use something that doesn't require installation.

PR #184, opened 2017-05-14
Allow only static final fields in classes annotated with @Event. (Fixes #182)

### LINKED ISSUE #182: error: Event fields must be declared as public non-final.

I'm trying to write a custom Event in my application, that includes `@IntDef` annotation, and contains `public final static int` fields.
I'm getting an error from the annotations processor that says 
`Error:(19, 9) error: Event fields must be declared as public non-final.`

## Expected Behavior
I would expect that the annoation processor skips static fields, and the build succeeds with no problem.

## Link to Code
Here is a sample code.
```
@Event
public class VoteEvent {
    public final static int VOTE_DIRECTION_UP = 1;
    public final static int VOTE_DIRECTION_DOWN = 2;

    @IntDef({VOTE_DIRECTION_UP, VOTE_DIRECTION_DOWN})
    @interface VoteDirection {}


    public @VoteDirection int voteDirection;
}

```
If you confirmed this as an issue, I'd be glad to submit a PR.

R4md4c:
I made two changes, the first change is allowing adding static final variables to the event class, and the second one is skipping the generation of final fields assignments in the generated dispatch method.

### DIFF
--- a/litho-processor/src/main/java/com/facebook/litho/specmodels/generator/EventGenerator.java
+++ b/litho-processor/src/main/java/com/facebook/litho/specmodels/generator/EventGenerator.java
@@ -9,17 +9,14 @@
 
 package com.facebook.litho.specmodels.generator;
 
-import javax.lang.model.element.Modifier;
-
 import com.facebook.litho.annotations.FromEvent;
 import com.facebook.litho.annotations.Param;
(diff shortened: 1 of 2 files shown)


IanChilds:
Thanks for doing this!

PR #188, opened 2017-05-22
Updating the Java Example for Solid Color.

adiphos:
The text above mentions that SolidColor component that takes a `colorRes` prop..
In the actual Java Sample a different property is mentioned.


passy:
Good catch. Thanks, @adiphos!

PR #192, opened 2017-05-25
[WIP] Upgrade to Gradle Plugin 3.0

passy:
This is a work-in-progress for upgrading to the new gradle plugin. It
significantly improves build times as it no longer implicitly propagates
`compile` dependencies and doesn't run all annotation processors it can
find on the class path.

It should also allow us to get rid of the `RELEASE_MODE` hack as it
properly supports variant propagation.

Sadly, the tests are currently failing due to this error:

```
java.lang.IllegalArgumentException: The key must be an application-specific resource id.
	at android.view.View.setTag(View.java:14978)
	at com.facebook.litho.ComponentHost.setTag(ComponentHost.java:379)
```

My suspicion is that this is because of Robolectric and not something we
do, but I'd be happy to be proven wrong about this.

I'll leave this around here for comments and if someone has a bunch on
how to fix the resource lookup, please let me know!

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,9 +1,10 @@
 buildscript {
     repositories {
         jcenter()
+        maven { url 'https://maven.google.com/' }
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.3.1'
+        classpath 'com.android.tools.build:gradle:3.0.0-alpha1'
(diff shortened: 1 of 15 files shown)


passy:
Update: Just tested this with Robolectric 3.3.2 and the issue still happens.

PR #194, opened 2017-06-01
[WIP] Upgrade to Gradle Plugin 3.0

passy:
Replaces #192 

This is a work-in-progress for upgrading to the new gradle plugin. It
significantly improves build times as it no longer implicitly propagates
`compile` dependencies and doesn't run all annotation processors it can
find on the class path.

It should also allow us to get rid of the `RELEASE_MODE` hack as it
properly supports variant propagation.

Sadly, the tests are currently failing due to this error:

```
java.lang.IllegalArgumentException: The key must be an application-specific resource id.
	at android.view.View.setTag(View.java:14978)
	at com.facebook.litho.ComponentHost.setTag(ComponentHost.java:379)
```

I've raised this with the Google tools folks and they're working on something to address this. The short version is that the resource ID generation has changed and is now supposed to go through AAPT for the final assignment, while Robolectric tries to use them as-is. Since the IDs are very low, it will fail the test in `View.java` and bail.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,9 +1,10 @@
 buildscript {
     repositories {
         jcenter()
+        maven { url 'https://maven.google.com/' }
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.3.3'
+        classpath 'com.android.tools.build:gradle:3.0.0-alpha4'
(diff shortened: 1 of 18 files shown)


passy:
This is now on `alpha4` which should hopefully address all our issues.

PR #208, opened 2017-07-03
TUTORIAL_DOC_UPDATE: 

vkalmath:
RecyclerBinder sample in tutorial is using builder which is not working in latest litho version 0.3.1. RecyclerBinder Constructors are public and builder method is not present.

### DIFF
--- a/docs/_docs/tutorial.md
+++ b/docs/_docs/tutorial.md
@@ -132,9 +132,9 @@ In this part of the tutorial, you'll use a [RecyclerBinder](/javadoc/com/faceboo
 First, in your activity, modify the `Component` definition as follows:
 
 ```java
-final RecyclerBinder recyclerBinder = new RecyclerBinder.Builder()
-    .layoutInfo(new LinearLayoutInfo(this, OrientationHelper.VERTICAL, false))
-    .build(context);
+final ComponentContext context = new ComponentContext(this);
+
+final RecyclerBinder recyclerBinder = new RecyclerBinder(context, new LinearLayoutInfo(this, OrientationHelper.VERTICAL, false));
 
 final Component component = Recycler.create(context)
     .binder(recyclerBinder)


passy:
Sorry, could you explain what you mean with "is not working"? It seems like the documentation is in line with what we use in the sample app: https://github.com/facebook/litho/blob/de4394d4127577da50f23738f9116c09e43af61e/sample/src/main/java/com/facebook/samples/litho/DemoListComponentSpec.java#L28

vkalmath:
Builder method is not present in [litho-widget-0.3.1] version. Only public
constructors are present. Attached the screenshots for reference.[image:
Inline image 1][image: Inline image 2]

passy:
Sorry, the attached images did not work. Could you please upload them here in the issue?

The `Builder` is exported as public as you can see here: https://github.com/facebook/litho/blob/61fa97ca7238c27d68d9db7f170e136f8c362fe0/litho-widget/src/main/java/com/facebook/litho/widget/RecyclerBinder.java#L144

vkalmath:
symbol not found
working with constructor


RecyclerBinder.java Source file from litho-widget-0.3.1-sources.jar also attached which does not have public Constructor Builder.
[RecyclerBinder.txt](https://github.com/facebook/litho/files/1125647/RecyclerBinder.txt)

passy:
@vgaidarji I managed to reproduce this now by depending on the published artifacts rather than the code. I'm not quite sure how this happened, but I'm looking into it! Thanks for the report!

vgaidarji:
@passy probably you meant someone else and not me as I wasn't involved in this issue :)

passy:
@vgaidarji Apologies, I did indeed. :)

passy:
I should have looked at the history earlier. This was only added a few weeks ago:

https://github.com/facebook/litho/commit/19ce14e31a66662de472f94d2a9abe70e86b1a66

So if you want to use that new API, I'd recommend using the `SNAPSHOT` release of the library. It will be included in the next stable release.

PR #225, opened 2017-08-14
litho-annotation should be a provided_deps

jollycopper:
Java files with litho annotations needs to be compiled with a "provided_deps" of litho-annotation. Putting litho-annotation in "deps" will cause duplicate jar inclusion when building the APK.

Note "provided_deps" is not documented in android_library's APIs.

@passy PTAL :)


passy:
Awesome, thank you!

PR #227, opened 2017-08-19
Enable support for change animations in ItemAnimator

BartNijland91:
What I wanted to achieve in an app in which I'm using Litho, is a crossfade animation between the old view and the new view after updating a component in the `Recycler`.

ItemAnimators have support for this, you can override the ``canReuseUpdatedViewHolder`` method inside an ``ItemAnimator`` and, when returning false, will enable both views (the old and the new one) to be alive at the same time when doing an update of a view on a certain position.

So for a demo I wrote inside the Playground a simple``Recycler`` which holds some components (rectangles with black color). I then update the items to a component with a cyan color with the ``updateItemAt`` method in the ``RecyclerBinder`` . I wrote the custom ItemAnimator and in the callback where you get both views, I set the correct alpha's and start the crossfade animation. This was the result:

litho_itemanimator_change_wrong

No animation is triggered. While debugging I noticed that both views where the same new view, and not the old one and the new. I digged around a little in the ``RecyclerBinder`` and noticed that when updating a component, it does not actually create a new one but modifies the existing one. When replacing that code with creating a new  ``ComponentTree`` it magically works.

litho_itemanimator_change_correct

I did not dig that deep into the source code and I do not know if this would be the proper solution, but at least this fixes the issue for me.

### DIFF
--- a/litho-widget/src/main/java/com/facebook/litho/widget/RecyclerBinder.java
+++ b/litho-widget/src/main/java/com/facebook/litho/widget/RecyclerBinder.java
@@ -480,7 +480,14 @@ public class RecyclerBinder
       if (previousRenderInfo.rendersView()) {
         updateViewCreatorMappingsOnItemRemove(previousRenderInfo);
       }
-      holder.setRenderInfo(renderInfo);
+      final ComponentTreeHolder newHolder = mComponentTreeHolderFactory.create(
+              renderInfo,
+              mLayoutHandlerFactory != null ?
+                      mLayoutHandlerFactory.createLayoutCalculationHandler(renderInfo) :
+                      null,
+              mCanPrefetchDisplayLists,
+              mCanCacheDrawingDisplayLists);
+      mComponentTreeHolders.set(position, newHolder);
       if (renderInfo.rendersView()) {
         updateViewCreatorMappingsOnItemInsert(renderInfo);
       }


pasqualeanatriello:
Hey @ScrollBart thanks for the PR! Unfortunately creating a new ComponentTree for an update operation is not going to be an efficient (nor correct) solution. It would miss out on previous state from the same row and it would also prevent any diffing on the layout.
The reason why we don't support change animations with ItemAnimator, is that you should be able to achieve the same with our internal Animation API.
cc @astreet that could help you more with how you can use the Animation APIs to achieve the crossfade effect.

ScrollBart:
@astreet Could you explain how to explain how to achieve this using the internal Animation API?

astreet:
@ScrollBart it's probably not going to be as easy as it was made to sound right now, as we're still working on that animations API and haven't actually open sourced any of the documentation on it yet as it's still being adopted within Facebook.

If you're interested in digging in, you can check out the Transition class and documentation there, and the classes under https://github.com/facebook/litho/tree/master/sample/src/main/java/com/facebook/samples/litho/transitionsdemo in the sample app. Right now, we support animating X/Y/opacity/scale. You'd have to implement an AnimatedProperty for color as well (see the AnimatedProperties class).

Unfortunately this isn't super streamlined right now and I can't guarantee you won't run into issues :/

ScrollBart:
@astreet Hmm.. what i'm trying to achieve here is to transition from a loading state UI to the actual content UI. Both are cards but with different UI's in them. The reason why I like to use different components for this (or different viewholders in "native" android) is that the UI and logic is encapsulated in separate objects, let's say LoadingComponent and CardComponent. That means that if I do an update operation, scroll all the way down an back again it would only need to render and know about the CardComponent.

Another thing is that on initial startup I add all these loading components. Then I start lazy loading the content and in when the data call is done, I can easily just say; update an component to a new one at that position and I'm done. I don't see how I could achieve the same thing here with the Litho Animation API, maybe make the component have a state prop (loading,complete) that will trigger a transition once changed?

astreet:
I understand what you're saying, and my idea would probably be to have a component that encapsulates and decides to render between LoadingComponent/CardComponent based on whether you're loading or not, and then define the animation there.

If you do want to only render one or the other as you're currently doing, you can still define a transition in the CardComponent and then use the same transition key in the CardComponent and the LoadingComponent for the view you want to transition color: you just need to make sure you're setting the CardComponent on the existing ComponentTree and not creating a new ComponentTree.

As a bit of an explanation, these transitions are triggered on every (non-incremental) mount, which is what happens when you update props, update state, or set a new component on a ComponentTree. The current components in the new mount have their `onCreateTransition` lifecycle method called, which defines what transition keys we want to animate. We then compare the existence and mounted properties (e.g. X, Y, etc) and do an animated transition for all those properties on the mounted content with those keys.

Hope this makes sense, I'm sorry we don't have better docs/examples yet.

astreet:
Also, based on your other PR: I think it does kind of suck to have to use two different animation abstractions for appearing/disappearing rows vs changing rows. I don't blame you for wanting to just use ItemAnimator for both, so it might still be worth the effort to make work for you.

The issue is that ItemAnimator change animations crossfade between two views, while a ComponentTree can only be associated with a single view. This means that you'll need to clone the ComponentTree and apply the StateHandler to it so that it has the right state. Layout diffing is not super important to keep around I don't think. @pasqualeanatriello can probably give a bit more detail as we just talked in person about this.

ScrollBart:
@astreet @pasqualeanatriello So this would basically mean I can use my original approach but make sure it has the right state?

astreet:
And clone the component tree as well (you can do this by creating a new component tree that has a component of the same type with the same props, and then applying the state of the other component).

ScrollBart:
👍 Can you explain how that differs from what I'm doing right now? Why should you reuse the old component and set a new state to it, while it is a completely different component (loading and content loaded cards/component)

astreet:
Closing old PR without response

PR #228, opened 2017-08-19
Set component key inside LithoView

BartNijland91:
When doing view animations inside an ``ItemAnimator`` you would typically check the ``ViewHolder`` or ``View`` instance to check what kind of view you are animating for. Since Litho uses the same classes (``BaseViewHolder`` and ``LithoView``) you cannot really do that. 

This PR adds a ``mComponentKey`` property to the ``LithoView`` so you can do your animations based on what type of ``Component`` you are dealing with.


passy:
cc @astreet @marco-cova

astreet:
Discussion is over at https://github.com/facebook/litho/pull/227

ScrollBart:
This is still relevant here based on the add and remove operations you can do on a `ItemAnimator` right?

astreet:
Closing along with #227

Issue #203, opened 2017-06-21
recyclerBinder.append()


Zeyad-37:
- [x] I have searched [existing issues](https://github.com/facebook/litho/issues) and this is not a duplicate

## Issues and Steps to Reproduce
Add an append function to add items to the list

## Expected Behavior
recyclerBinder.append(/*item(s) to added*/)

passy:
@Zeyad-37 What would that add that the existing functions don't provide? E.g. `insertItemAt`, `updateItemAt` etc.?

Zeyad-37:
That i dont need to manage the index

passy:
I've now noticed this being quite inconvenient at times too. If someone wants to prepare a PR for this, I'm happy to review it. :)

mihaelao:
closed in #236

PR #236, opened 2017-09-05
Add appendItem for RecyclerBinder

leochuan:
Pull request for #203

### DIFF
--- a/docs/_docs/recycler-component.md
+++ b/docs/_docs/recycler-component.md
@@ -58,6 +58,7 @@ final RecyclerBinder recyclerBinder = new RecyclerBinder.Builder()
 The most commonly used are:
 
 ``` java
+recyclerBinder.appendItem(component)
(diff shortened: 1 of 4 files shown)


passy:
Thanks for your contribution!

Could you add some tests for this too? :) The proper suite for this would be 
https://github.com/facebook/litho/blob/master/litho-it/src/test/java/com/facebook/litho/widget/RecyclerBinderTest.java

leochuan:
@passy ok :)

leochuan:
@passy It's done, cause `appendItem` internally call the method `insertItemAt` so I think test whether it insert items at tail is enough.

passy:
Thanks a lot! Yeah, that sounds sensible. I was thinking about potential issues with the runtime complexity here, but `getItemCount()` should be O(1) as long as we use an `ArrayList` to hold on to the items.

Issue #206, opened 2017-06-27
Support `resType` for var-arg properties


passy:
For var-arg properties, we want to support `resType` constraints for the provided arguments in the same way that you currently can for regular props.

E.g.

```java
@OnCreateLayout
static ComponentLayout onCreateLayout(
   LayoutContext context,
   @Prop(varArg = "text", resType = ResType.STRING) List<CharSequence> textBlocks) {
   ...
}
```

*Point of contact:* @passy

passy:
Fixed by #240

PR #240, opened 2017-09-14
Support  for var-arg properties

leochuan:
PR for #206 
Cause `varArgBuilder` and `regularBuilder` handle the default generating properly. 
So I reconstruct `attrBuilders` `resBuilder` `pxBuilder` `dipBuilder` and `sipBuilder` for supporting this feature. 

There are two builders, `resTypeListBuilder` for generating method with `List` parameter, `resTypeRegularBuilder` for generating method with normal parameter. 
The code generated looks like below:
```java
public Builder sRes(@DimenRes int resId) {
  if (this.mPlaygroundComponentImpl.s == null) {
    this.mPlaygroundComponentImpl.s = new ArrayList<Float>();
  }
  final float res = resolveFloatRes(resId);
  this.mPlaygroundComponentImpl.s.add(res);
  mRequired.set(0);
  return this;
}

public Builder sRes(List<Integer> resIds) {
  if (resIds == null) {
    return this;
  }
  if (this.mPlaygroundComponentImpl.s == null) {
    this.mPlaygroundComponentImpl.s = new ArrayList<Float>();
  }
  for (int i = 0; i < resIds.size(); i++) {
    final float res = resolveFloatRes(resIds.get(i));
    this.mPlaygroundComponentImpl.s.add(res);
  }
  mRequired.set(0);
  return this;
}
```
Also, If the parameter type is something like `List<Float>` I will unbox the Float,
 or `final Float res = resolveFloatRes(resIds.get(i))` will emit an error.

For `ResType.INT_ARRAY`, `int[]` can't be auto-box into `Integer[]` , so I add two methods to return `Integer[]` in `ResourceResolver.java`. And change the modifier of `resolveStringArrayRes` from `private` to `protected` or it will emit an error when resType is `ResType.STRING` 

And I find that if `resType` is `DIMEN` and parameter is `Float`. `this.mPlaygroundComponentImpl.f = resolveDimenSizeRes(resId);` will emit an error. I fix that by changing generated code of `resType.non-null` like this:
```java
public Builder fRes(@DimenRes int resId) {
  this.mPlaygroundComponentImpl.f = (float) resolveDimenSizeRes(resId);
  mRequired.set(0);
  return this;
}
```
I also handle the `xPx(@Px float f)` method properly, since it is a little bit different from others.
Finally, I remove the warning in `PropValidation.java` of using both `varArgs` and `resType`, and add some new validations.

### DIFF
--- a/docs/_docs/props.md
+++ b/docs/_docs/props.md
@@ -167,15 +167,15 @@ MyComponent.create(c)
    .name("Three")
 ```
 
-If you specific `resType`:
+You can also combine `varArg` with `resType` props:
 
 ```java
 MyComponent.create(c)
-   .namePx(1f)
-   .nameRes(resId)
-   .nameAttr(attrResId)
-   .nameDip(1f)
-   .nameSp(1f)
+   .textSizePx(1f)
+   .textSizeRes(resId)
+   .textSizeAttr(attrResId)
+   .textSizeDip(1f)
+   .textSizeSp(1f)
 ```
 
 
--- a/docs/_docs/props.md
+++ b/docs/_docs/props.md
@@ -167,7 +167,18 @@ MyComponent.create(c)
    .name("Three")
 ```
 
-Note that, when using `varArg`, `resType`s are not supported.
+If you specific `resType`:
+
+```java
+MyComponent.create(c)
+   .namePx(1f)
+   .nameRes(resId)
+   .nameAttr(attrResId)
+   .nameDip(1f)
+   .nameSp(1f)
+```
+
+
 
 ## Immutability
 The props of a Component are read-only. The Component's parent passes down values for the props when it creates the Component and they cannot change throughout the lifecycle of the Component. If the props values must be updated, the parent has to create a new Component and pass down new values for the props.
--- a/litho-it/src/test/java/com/facebook/litho/specmodels/generator/BuilderGeneratorTest.java
+++ b/litho-it/src/test/java/com/facebook/litho/specmodels/generator/BuilderGeneratorTest.java
@@ -63,13 +63,23 @@ public class BuilderGeneratorTest {
     public void testUpdateStateMethod() {}
   }
 
+  @LayoutSpec
+  static class TestResTypeWithVarArgsSpec {
+    @OnCreateLayout
+    public void test(@Prop(varArg = "test", resType = ResType.DIMEN_TEXT) List<Float> arg0) {}
+  }
+
   private SpecModel mSpecModel;
+  private SpecModel mSpecModel1;
 
   @Before
   public void setUp() {
     Elements elements = mCompilationRule.getElements();
     TypeElement typeElement = elements.getTypeElement(TestSpec.class.getCanonicalName());
     mSpecModel = mLayoutSpecModelFactory.create(elements, typeElement, null);
+
+    TypeElement typeElement1 = elements.getTypeElement(TestResTypeWithVarArgsSpec.class.getCanonicalName());
+    mSpecModel1 = mLayoutSpecModelFactory.create(elements, typeElement1, null);
   }
 
   @Test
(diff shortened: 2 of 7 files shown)


emilsjolander:
This is awesome! I'll let others review the code but I would really appreciate it if you could also include an edit to the docs which currently say that restype is not supported (https://fblitho.com/docs/props#variable-arguments).

passy:
Thanks so much for doing this! This looks great! I'll give a more thorough review later today. For the docs, that's also fine to do in a separate commit or PR if you prefer.

leochuan:
@emilsjolander @passy The test failed cause the generated code is different from `TestMount.java`.It turns up my `TestResTypeWithBoxType` is redundant. Maybe I can update `TestMount.java` and the docs after reviewing

docs/_docs/props.md
```diff
@@ -167,7 +167,18 @@ MyComponent.create(c)
    .name("Three")
 ```
 
-Note that, when using `varArg`, `resType`s are not supported.
+If you specific `resType`:
+
+```java
+MyComponent.create(c)
+   .namePx(1f)
```
passy: `name` is probably not a great example here as it's a string resource. Could you add a spec with a resource type to illustrate this as above?

litho-it/src/test/java/com/facebook/litho/specmodels/generator/BuilderGeneratorTest.java
```diff
@@ -63,13 +63,23 @@ public void testEventMethod(@Prop boolean arg0) {}
     public void testUpdateStateMethod() {}
   }
 
+  @LayoutSpec
+  static class TestResTypeWithVarArgsSpec {
+    @OnCreateLayout
+    public void test(@Prop(varArg = "test", resType = ResType.DIMEN_TEXT) List<Float> arg0) {}
```
passy: Could you use some more descriptive names here? `test` twice makes this a bit difficult to follow.

litho-it/src/test/java/com/facebook/litho/specmodels/generator/BuilderGeneratorTest.java
```diff
@@ -186,4 +196,216 @@ public void testGenerate() {
...
+            "\n" +
+            "  private void init(com.facebook.litho.ComponentContext context, int defStyleAttr, int defStyleRes,\n" +
+            "      TestResTypeWithVarArgsImpl testResTypeWithVarArgsImpl) {\n" +
+            "    super.init(context, defStyleAttr, defStyleRes, testResTypeWithVarArgsImpl);\n" +
+            "    mTestResTypeWithVarArgsImpl = testResTypeWithVarArgsImpl;\n" +
+            "    mContext = context;\n" +
+            "    mRequired.clear();\n" +
+            "  }\n" +
+            "\n" +
+            "  public Builder test(java.lang.Float test) {\n" +
```
passy: Sorry, I may just be missing something here, but is this needed? Seems to be the same as below apart from the boxing.

litho-it/src/test/java/com/facebook/litho/specmodels/model/PropValidationTest.java
```diff
@@ -210,6 +210,20 @@ public void testVarArgPropMustHaveParameterizedListType() {
...
+  public void testIncorrectTypeForResTypeWithVarArg() {
+    when(mPropModel1.getResType()).thenReturn(ResType.BOOL);
+    when(mPropModel1.hasVarArgs()).thenReturn(true);
+    when(mPropModel1.getType()).thenReturn(ParameterizedTypeName.get(ClassNames.LIST, TypeName.INT.box()));
+
+    List<SpecModelValidationError> validationErrors = PropValidation.validate(mSpecModel, PropValidation.RESERVED_PROP_NAMES);
+    assertThat(validationErrors).hasSize(1);
+    assertThat(validationErrors.get(0).element).isEqualTo(mRepresentedObject1);
+    assertThat(validationErrors.get(0).message).isEqualTo(
+        "A prop declared with resType BOOL must be one of the following types: " +
```
passy: Could you add a note to the error message that this is about a var-arg prop?

litho-it/src/test/resources/processor/TestMount.java
```diff
@@ -634,25 +634,29 @@ public Builder prop7(CharSequence prop7) {
     }
 
     public Builder prop7Res(@StringRes int resId) {
-      this.mTestMountImpl.prop7 = resolveStringRes(resId);
+      final CharSequence res = resolveStringRes(resId);
+      this.mTestMountImpl.prop7 = res;
```
passy: Would it be possible to collapse the lines here?

litho-it/src/test/java/com/facebook/litho/specmodels/generator/BuilderGeneratorTest.java
leochuan: I was thinking of this neither. To be honest, I don't know why should I need a `Float` method with some `resType`s, But maybe when you declaring `varArgs` with `List<Float>` generate a method with `Float` is a right behaviour. I am not sure about that or I can just remove it.

litho-it/src/test/resources/processor/TestMount.java
leochuan: I did this because `DIMEN_SIZE` `DIMEN_TEXT` and `DIMEN_OFFSET` can take both `Integer` and `Float` but `resolveDimenSize` only return `int`.So when you specific resType as any of them and pass a `Float ` there will be an error. 

I can collapse the lines by adding a cast when there is `Float` type with a `resolveDimenSize` method so that I should add `dimenResBuilders` and `dimenAttrBuilders` to handle this. And Also I will add a test.

docs/_docs/props.md
```diff
@@ -167,7 +167,18 @@ MyComponent.create(c)
    .name("Three")
 ```
 
-Note that, when using `varArg`, `resType`s are not supported.
+If you specific `resType`:
```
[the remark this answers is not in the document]
leochuan: It's done.

docs/_docs/props.md
leochuan: It's done I replace it with `textSize`

litho-it/src/test/java/com/facebook/litho/specmodels/model/PropValidationTest.java
leochuan: It's done

litho-it/src/test/resources/processor/TestMount.java
leochuan: It's done

litho-it/src/test/java/com/facebook/litho/specmodels/generator/BuilderGeneratorTest.java
```diff
@@ -63,13 +63,33 @@ public void testEventMethod(@Prop boolean arg0) {}
...
+  private SpecModel mSpecModel2;
 
   @Before
   public void setUp() {
     Elements elements = mCompilationRule.getElements();
     TypeElement typeElement = elements.getTypeElement(TestSpec.class.getCanonicalName());
     mSpecModel = mLayoutSpecModelFactory.create(elements, typeElement, null);
+
+    TypeElement typeElement1 = elements.getTypeElement(TestResTypeWithVarArgsSpec.class.getCanonicalName());
+    mSpecModel1 = mLayoutSpecModelFactory.create(elements, typeElement1, null);
```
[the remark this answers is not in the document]
leochuan: I looked through other files and find that `1 2 3` pattern..Sometimes it's really hard to pick a name, now the name is `mResTypeVarArgsSpecModel ` :)

litho-it/src/test/java/com/facebook/litho/specmodels/generator/BuilderGeneratorTest.java
leochuan: Now the `varArg` names `size` and the argument names `sizes` maybe it's easier to follow

litho-it/src/test/java/com/facebook/litho/specmodels/generator/BuilderGeneratorTest.java
leochuan: Since `DIMEN_SIZE` `DIMEN_OFFSET` and `DIMEN_TEXT` do not add regularBuilder so I figure out that the `Float` method is indeed redundant. I move the `varArgBuilder` block into `ResType.None` and add a new `regularBuilders` to handle both `regularBuilder` and `varArgBuilder` for `non-dimen` `resType` so it's just as same as before in concept.

leochuan:
@passy I've fixed all review changes

litho-it/src/test/java/com/facebook/litho/specmodels/generator/BuilderGeneratorTest.java
passy: Awesome!

passy:
@leochuan Thanks so much for the updates! I've imported this again and will let some internal people take a look. That probably won't happen before Monday, though.

passy:
Landed! Thank you so much, @leochuan!

PR #241, opened 2017-09-16
Update recycler-component.md

tatocaster:
after the 0.6.0 `ComponentInfo` renamed to `RenderInfo`, so `ComponentRenderInfo` class is required in this example.


passy:
Thank you!

PR #256, opened 2017-10-26
Update using-components.md

TGIO:
minor doc fix


zpao:
Sorry for stealing the import :) There was a problem with our tooling during the previous import and I was confirming that it still worked after making a config change.

PR #261, opened 2017-10-29
Update link images

kant:


kant:
Just added the `/docs/` folder to the path. So, I fixed to see the images.

passy:
@kant I'm not sure why you don't see the images, but they show up just fine here: https://fblitho.com/docs/debugging

PR #264, opened 2017-10-30
Update the image's links

kant:
Added `/docs` to the path to allow to see the original images.


IanChilds:
This is the same as #266

PR #268, opened 2017-11-02
Delay variable initialization

pavlospt:
`propDefaultResType` & `propDefaultResId` were initialized before their actual usage. After their initialization, `extractFromMethod()` could still return without using them.


passy:
That was so fast!

PR #271, opened 2017-11-02
Formatting and typo

Pavneet-Sing:


dsyang:
Still looks like there's 4 commits :( @passy  does that matter when importing?

Pavneet-Sing:
@dsyang yeah it is 4 oops, i thought you were saying that because of that failed check that was occurring due to one commit from a patch

passy:
@dsyang Doesn't matter when importing.

Thanks, but this actually doesn't match up with our internal style guide. We follow the [Google Java Styleguide](https://google.github.io/styleguide/javaguide.html) and while we're not 100% conformant right now, we want to try to get at least all code we touch to follow it from now on. :)

Pavneet-Sing:
You mean i violated this "Tab characters are not used for indentation" , sometimes it's kinda hard when you have to choose between space and tab

passy:
@Pavneet-Sing In particular the alignment of assignments or similar things. This may look nicer at a glance but creates issues in code reviews as the addition of a single item can cause everything else to reflow, adding irrelevant line changes to the review. That's why most guidelines for larger codebases don't use it.

Pavneet-Sing:
@passy Thank you for taking the time to enlighten me

passy:
@Pavneet-Sing No problem. Thanks for the pull request. Keep them coming! :)

PR #275, opened 2017-11-16
Docs: Fix broken link to debugHighlightMountBounds

drd:
I have previously signed the Facebook CLA, but to be clear I have not tested this by actually generating the docs from this. Please let me know if you'd like me to do so.

This fixes a broken link from when Dev Options page was created: https://github.com/facebook/litho/commit/a1207a8ce7227ebd6107ea5e17a420edd3d58c67#diff-120012bbe26d93715d4f73fb9328e38c


passy:
Good catch! Thank you!

PR #297, opened 2018-01-04
Tutorial: return ComponentLayout from ListItemSpec#onCreateLayout()

Saketme:
onCreateLayout() had a return type of Component, which is incorrect.


IanChilds:
Hi @Saketme, thanks for the contribution. However, if you look at the latest code in master, you'll see that this is a recent change that we made, so the docs are actually correct.

Saketme:
Ah, sorry for this PR then. Carry on :)

PR #304, opened 2018-01-23
Update tutorial.md

Dragon-Boat:
change "Component" to "ComponentLayout".

### DIFF
--- a/docs/_docs/tutorial.md
+++ b/docs/_docs/tutorial.md
@@ -84,7 +84,7 @@ Your custom component will be called `ListItem` and it will display a title with
 public class ListItemSpec {
 
   @OnCreateLayout
-  static Component onCreateLayout(ComponentContext c) {
+  static ComponentLayout onCreateLayout(ComponentContext c) {
 
     return Column.create(c)
         .paddingDip(ALL, 16)


IanChilds:
Hi @Dragon-Boat, thanks for the pull request! However, the docs do actually match the latest master.

Dragon-Boat:
`@LayoutSpec
public class ListItemSpec {

  @OnCreateLayout
  static Component onCreateLayout(ComponentContext c) {

    return Column.create(c)
        .paddingDip(ALL, 16)
        .backgroundColor(Color.WHITE)
        .child(
            Text.create(c)
                .text("Hello world")
                .textSizeSp(40))
        .child(
            Text.create(c)
                .text("Litho tutorial")
                .textSizeSp(20))
        .build();
  }
}`
i typed this code as the "Tutorial".  the onCreateLayout method return a Component but the Column.create().build() return a ComponentLayout. so the AS gives the errorMessage " ComponentLayout can't be converted to Component

IanChilds:
In the latest master, `Column` is a `Component`. If you're using the latest release, rather than the latest snapshot release, then you might run into the problem that you're talking about here.

PR #325, opened 2018-03-12
Update README.md

lauraguevara97:
added BSD License icon :)


passy:
Hey Laura,

thanks for the contribution, but we use a 3-clause BSD license with an additional patents grant. There's no dedicated page for that variant on opensource.org, so I'd rather not send people there and get the wrong impression about the actual licensing situation. If you find something else in the README (or elsewhere!) that could be improved, please send it along. :)

PR #326, opened 2018-03-12
Update broken links

pavlospt:


pavlospt:
@passy they seem to be broken when navigating from this page https://github.com/facebook/litho/blob/master/docs/_docs/component-tree.md

pavlospt:
@passy Hmm, interesting...you can close this then :)

PR #344, opened 2018-04-17
Transfer external Annotations on Component fields

### LINKED ISSUE #343: Transfer @Nullable annotation from @Prop to generated Component
When marking a `@Prop` as `@Nullable`, the generated component should have the field and the parameter in the builder method also generated with `@Nullable`. This helps static analysis.

Point of contact: @passy

pavlospt:
* Builder methods were already including external Annotations
for each Prop. Followed a same approach to transfer them on the Component
fields as well.

Closes #343


passy:
Sweet! Could you perhaps modify one of our integration tests to include a `@Nullable` field to see how the change affects the generated code?

passy:
This one might be a good candidate: https://github.com/facebook/litho/blob/master/litho-it/src/test/resources/processor/TestLayoutSpec.java

pavlospt:
Sure will update the PR with a test case for it!

pavlospt:
@passy added 2 tests cases: 

- `ProcessorIntegrationTest.java`
- `ComponentBodyGeneratorTest.java`

passy:
A dream! Thank you. I'll try and get that through internal review quickly. :)

pavlospt:
@passy let me check the CI failure!

passy:
@pavlospt Looks like there's a missing buck dep on the jsr annotations.

pavlospt:
@passy yeap found that as well! Still figuring out `BUCK` but it was easier than last time 😂

PR #347, opened 2018-04-24
Add Component.Builder variant method

### LINKED ISSUE #89: Add .component(Component.Builder) variant to ComponentInfo.Builder
So that we don't force users to explicitly call build() when updating a `RecyclerBinder`.

Point of contact: @pasqualeanatriello

pavlospt:
* Adds method in `ComponentRenderInfo.Builder` that accepts a `Component.Builder`.

Closes #89

### DIFF
--- a/litho-widget/src/main/java/com/facebook/litho/widget/ComponentRenderInfo.java
+++ b/litho-widget/src/main/java/com/facebook/litho/widget/ComponentRenderInfo.java
@@ -76,6 +76,11 @@ public class ComponentRenderInfo extends RenderInfo {
       return this;
     }
 
+    public Builder component(Component.Builder builder) {
+      this.component = builder.build();
+      return this;
+    }
+
     public ComponentRenderInfo build() {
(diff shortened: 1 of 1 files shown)


litho-widget/src/main/java/com/facebook/litho/widget/ComponentRenderInfo.java
```diff
@@ -76,6 +76,11 @@ public Builder component(Component component) {
       return this;
     }
 
+    public Builder component(Component.Builder builder) {
+      this.component = builder.build();
```
IanChilds: I think a more common pattern here would just to do `return component(builder.build());`, what do you think? 

Thanks for doing this!

pavlospt: @IanChilds sure that would save possible issues in case more things are added to `.component(Component component)` method! Will change it!

Issue #348, opened 2018-04-25
User can't make elevation 0


mishrabhilash:
## Issues
Minimum elevation of card widget is 2dp. If user provides zero elevation, application crashes with Illegal state exception

```
Process: com.fblitho.lithoktsample, PID: 11024
    java.lang.IllegalArgumentException: radius must be > 0
        at android.graphics.RadialGradient.<init>(RadialGradient.java:61)
        at com.facebook.litho.widget.CardShadowDrawable.buildShadow(CardShadowDrawable.java:179)
        at com.facebook.litho.widget.CardShadowDrawable.draw(CardShadowDrawable.java:94)
        at com.facebook.litho.ComponentHost$InterleavedDispatchDraw.drawNext(ComponentHost.java:1234)
        at com.facebook.litho.ComponentHost$InterleavedDispatchDraw.access$300(ComponentHost.java:1186)
        at com.facebook.litho.ComponentHost.dispatchDraw(ComponentHost.java:652)
        at android.view.View.draw(View.java:19985)
        at android.view.View.updateDisplayListIfDirty(View.java:18857)
        at android.view.View.draw(View.java:19710)
        at android.view.ViewGroup.drawChild(ViewGroup.java:4321)
        at android.support.v7.widget.RecyclerView.drawChild(RecyclerView.java:4581)
        at android.view.ViewGroup.dispatchDraw(ViewGroup.java:4107)
        at android.view.View.draw(View.java:19985)
        at android.support.v7.widget.RecyclerView.draw(RecyclerView.java:3987)
        at android.view.View.updateDisplayListIfDirty(View.java:18857)
        at android.view.View.draw(View.java:19710)
        at android.view.ViewGroup.drawChild(ViewGroup.java:4321)
        at android.view.ViewGroup.dispatchDraw(ViewGroup.java:4107)
        at android.view.View.draw(View.java:19985)
        at android.view.View.updateDisplayListIfDirty(View.java:18857)
        at android.view.View.draw(View.java:19710)
        at android.view.ViewGroup.drawChild(ViewGroup.java:4321)
        at android.view.ViewGroup.dispatchDraw(ViewGroup.java:4107)
        at com.facebook.litho.ComponentHost.dispatchDraw(ComponentHost.java:645)
        at android.view.View.draw(View.java:19985)
        at android.view.View.updateDisplayListIfDirty(View.java:18857)
        at android.view.View.draw(View.java:19710)
[...]

PR #349, opened 2018-04-25
Added check for zero elevation

mishrabhilash:
A solution for issue https://github.com/facebook/litho/issues/348
Earlier, if the user added zero elevation, the app was crashing. Now it is solved.

### DIFF
--- a/litho-widget/src/main/java/com/facebook/litho/widget/CardSpec.java
+++ b/litho-widget/src/main/java/com/facebook/litho/widget/CardSpec.java
@@ -89,32 +89,29 @@ class CardSpec {
     final int shadowBottom = getShadowBottom(elevation);
     final int shadowHorizontal = getShadowHorizontal(elevation);
 
-    Column.Builder builder = Column.create(c);
-
-        builder.child(
-                Column.create(c)
-                        .marginPx(HORIZONTAL, shadowHorizontal)
-                        .marginPx(TOP, shadowTop)
-                        .marginPx(BOTTOM, shadowBottom)
-                        .backgroundColor(cardBackgroundColor)
-                        .child(content)
-                        .child(
-                                CardClip.create(c)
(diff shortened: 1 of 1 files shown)


mishrabhilash:
I want to have a container with round edges but no elevation. I know this
can be solved with a drawable having rounded corners but Card with
customizable elevation seems a more general solution.

pavlospt:
Yeap but you might stumble upon other limitations of a `CardSpec`in future as well! Trying to use something specifically designed for another purpose, I am not sure if it can help you!

mishrabhilash:
That is how we will know about limitations and improve upon them :)

Moreover, Android CardView too supports zero elevation. So, I think litho
should as well.

It is not about how I intend to use it, rather, having zero elevation
support gives it a sense of completeness.

pavlospt:
Ok fair point :)

litho-widget/src/main/java/com/facebook/litho/widget/CardSpec.java
```diff
@@ -89,28 +89,32 @@ static Component onCreateLayout(
...
-                .backgroundColor(cardBackgroundColor)
-                .child(content)
-                .child(
-                    CardClip.create(c)
-                        .clippingColor(clippingColor)
-                        .cornerRadiusPx(cornerRadius)
-                        .positionType(ABSOLUTE)
-                        .positionPx(ALL, 0)))
-        .child(
-            CardShadow.create(c)
```
muraziz: You can inline that check in this line like below instead of breaking it down into builder:
```
     .child(
         elevation > 0 ? CardShadow.create(c)......
              : null)
```

mishrabhilash: applied inline check

passy:
Thank you, @mishrabhilash!

Issue #267, opened 2017-10-30
[meta-task] Kotlin outstanding issues


passy:
This issue is to keep track of remaining issues and where applicable work-arounds when using Litho with Kotlin.

## Issues

- [X] Cannot use Kotlin `object`s for specs.
- [X] Annotated methods inside specs require `@JvmStatic`.
- [X] `@PropDefault` doesn't work.
- [x] `@PropDefault` requires `@JvmField` annotation.

## Other Gotchas

- `kapt` is a pig. There seem to be a bunch of invalidation bugs that force you to clean your cache and rebuild.
- `NonExistentClass` shows up in random places. Again, this appears to be a `kapt` bug and the best workaround for now appears to be [`./gradlew clean`](https://stackoverflow.com/questions/45099688/error-error-nonexistentclass-kotlin-in-multi-module-dagger-project).

## Future work

We should consider generating Kotlin code to fully leverage the new language features like data classes, allow for Anko-like DSLs and get rid of builders.

pavlospt:
@passy `@PropDefault` does not work for which cases? Is there a branch with Kotlin sample available?

passy:
@pavlospt I'm working on fixing that today. I think Pasquale's Kotlin sample could use this in a spot.

pavlospt:
@passy Have you tried `@get:PropDefault val something = 1f` for example ?

passy:
@pavlospt Sadly, that generates even worse code for us at the moment. I've got a work-around ready with a long-ish explanation on the commit which should land probably tomorrow.

pavlospt:
Ok, great! Will wait to check that commit tomorrow :) Please ping it here as well in case I miss it :P

passy:
@pavlospt As if you'd ever miss a commit! 😜

pavlospt:
@passy My colleagues usually say the same thing, no wonder why 😂

yanex:
@passy What kind of bugs do you experience with kapt? If you can reproduce something going bad, please submit the issue to kotl.in/issue, and we will fix it shortly.

passy:
@yanex Hey! Thanks for commenting here. :)

Sadly, the main problem for us has exactly been reproducible. Everyone on our team who's worked with this has found themselves in a situation where compilation would fail for seemingly random reasons and only clearing the caches would help.

[...]

PR #362, opened 2018-05-19
Prop default without @JvmField annotation

pavlospt:
This PR generates a proper accessor for not `@JvmField` annotated `PropDefaults`

* Unfortunately this won't work if the `@PropDefault` field gets annotated with `@JvmField`. Did not find any workaround to it, since @JvmField annotation is not forwarded on `PropModel`'s annotations/external annotations.

* Addresses the last issue on #267


pavlospt:
One interesting observation while fixing this, is that if you use the `@PropDefault` like `@get:PropDefault` you can even save one extra synthetic method from the backing Java class.

passy:
Oh, that's amazing!

PR #368, opened 2018-05-23
Add Kotlin variant for MountSpec

pavlospt:
Per title


astreet:
Thanks for taking this on! However with this approach I'm concerned that we're copy-pasting docs and they can get out of sync. How do you feel about having kotlin and java examples in the same docs file? cc @passy do we have a pattern for java and kotlin docs yet?

pavlospt:
@astreet I think that having them in the same doc could end up in very long blocks of text. Also, I guess that it would mean more scrolling to find an example in your desired language. That's why I believe this approach could work better!

astreet:
Gotcha. Well, let's talk about this and figure out what we want to do before you make more of these PRs :) I would be opposed to landing something with duplicated docs, but maybe there's some other idea we can think of (e.g. a tabbed component that allows you to choose to show a block of code in different languages)

pavlospt:
Ohh okay, I thought that because the PR for the LayoutSpec landed, it was fine to open more PRs to that direction! 

The tabbed component you are talking about is similar to the buttons that exist right now If I am correct. It would be interesting to see if it can be nested inside the markdown code snippets. 

One thing I am worried about tabbing code snippets is that it might lead to maintaining a lot more content than two docs. 

e.g. Maintaining `props-java` & `props-kt` vs maintaining `props-java-snippetA` & `props-java-snippetB` + `props-kt-snippetA` & `props-kt-snippetB`

astreet:
Ah, I'm not sure why that one landed but I don't think it was a good idea -- I also don't think there's anything stopping us from inlining the snippets in the single doc page given we have the right component

pavlospt:
Okay, that makes sense! Should I wait for it to be discussed internally? :)

passy:
Given that the prose rarely requires any updates for Kotlin vs Java, it would be ideal if we could just keep buttons at the top to toggle the language of the embedded code snippets like Andy suggested.

pavlospt:
If that is ok I can try prototyping it :P Don't want to promise anything in advance though 😂

passy:
@pavlospt Of course! :) No problem if it doesn't lead anywhere.

pavlospt:
@passy @astreet Couldn't come up with something working as intended! I think someone more experienced from me on this stuff, can take it up :) I would be happy to modify any already opened PR and contribute even more Kotlin docs, once there is something ready!

passy:
I may give it a shot. Just don't know yet when I'll have time for it.

pavlospt:
Pinging this issue because I found an interesting component, similar to what we discussed being used in Gradle guides :) https://guides.gradle.org/performance/ Not sure if it is something that could be done!

passy:
Oh yeah, I like to look of that.

pavlospt:
@facebook-github-bot Nope this PR is not relevant any more, after the discussion above :) Closing!

PR #371, opened 2018-05-24
Fix PropDefault values extraction for Kotlin

pavlospt:
* There was an issue with the current implementation if a user tried to use that same annotation prefixed with `get`.

e.g.: `@get:PropDefault`

The solution is to identify which is the case and fallback to the respective method that holds the `@PropDefault` annotation on the equivalent Java file.

With `@get:`

```java
   @PropDefault(
      resType = ResType.DIMEN_SIZE,
      resId = 2131034112
   )
   @Nullable
   public final Float getImageAspectRatio() {
      return imageAspectRatio;
   }
```

Without `@get:`

```java
   /** @deprecated */
   // $FF: synthetic method
   @PropDefault(
      resType = ResType.DIMEN_SIZE,
      resId = 2131034112
   )
   public static void imageAspectRatio$annotations() {
   }
```

### DIFF
--- a/litho-it/src/test/java/com/facebook/litho/specmodels/processor/PropDefaultsExtractorTest.java
+++ b/litho-it/src/test/java/com/facebook/litho/specmodels/processor/PropDefaultsExtractorTest.java
@@ -18,6 +18,7 @@ package com.facebook.litho.specmodels.processor;
 import static org.assertj.core.api.Assertions.assertThat;
 
 import com.facebook.litho.annotations.PropDefault;
+import com.facebook.litho.annotations.ResType;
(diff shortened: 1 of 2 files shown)


passy:
Great stuff! As this is getting more complex, would it be possible to add a test case for this in `PropDefaultsExtractorTest`? I'm not quite sure if this would just be a matter of essentially copying over the two examples you have there or if there's a lot of additional setup that would be required.

pavlospt:
@passy we could copy the Java equivalent classes and use them as test cases!

passy:
Thanks!

Issue #373, opened 2018-05-25
varArg properties breaks when using most types.


kjbilliot:
- [x] I have searched [existing issues](https://github.com/facebook/litho/issues) and this is not a duplicate

## Version

`Config.Dependencies.litho` is `"0.14.0"`

```
    implementation("com.facebook.litho", "litho-core", Config.Dependencies.litho)
    implementation("com.facebook.litho", "litho-widget", Config.Dependencies.litho)
    compileOnly("com.facebook.litho", "litho-annotations", Config.Dependencies.litho)
    kapt("com.facebook.litho", "litho-annotations", Config.Dependencies.litho)
    kapt("com.facebook.litho", "litho-processor", Config.Dependencies.litho)
```

## Issues and Steps to Reproduce

1. Create either LayoutSpec or MountSpec.
2. Add `@Prop(varArg = "name") names : List<String>` as a property. 
3. Code should compile fine here.
4. Change types to something else outside of basic types (IE: Color, TextView, etc)
5. Code will no longer compile, giving the following two errors in the build panel, from the generated component .java file.
`error: illegal start of type    public Builder color(@NotNull) ? extends Color color) {`
`error: ';' expected    public Builder color(@NotNull) ? extends Color color) {`

## Expected Behavior
Code should compile fine, and I should be able to use the component in the way demonstrated in [the documentation](https://fblitho.com/docs/props#variable-arguments) for whatever type I'd need to be a varArg.

## Link to Code
Example code in Kotlin.

```
@LayoutSpec
object TestComponentSpec {
    @OnCreateLayout
    fun onCreateLayout(
        c : ComponentContext,
        @Prop(varArg = "color") colors : List<Color>
    ) : Component {
        return Row.create(c).build()
    }
}
```
Errors from Gradle:

```
e: /home/<snip>/android/app/build/generated/source/kapt/devDebug/com/mingo/chat/ui/TestComponent.java:130: error: illegal start of type
    public Builder color(@NotNull ? extends Color color) {
                                  ^
e: /home/<snip>/android/app/build/generated/source/kapt/devDebug/com/mingo/chat/ui/TestComponent.java:130: error: ')' expected
    public Builder color(@NotNull ? extends Color color) {
                                   ^
e: /home/<snip>/android/app/build/generated/source/kapt/devDebug/com/mingo/chat/ui/TestComponent.java:130: error: ';' expected
    public Builder color(@NotNull ? extends Color color) {
                                           ^
[...]

PR #374, opened 2018-05-25
Fix varArg builder methods for Kotlin

pavlospt:
Due to Kotlin compiler applying wildcards on List types, there was an issue when generating from a Kotlin spec and using `varArg` Props.

Easy fix was to annotate List type with `JvmSuppressWildcards` but just in case someone forgets to do it, this PR is taking care of handling it seamlessly, by removing wildcards iff generating from Kotlin specs.

Addresses #373


pavlospt:
Of cource, I think I can add some, but I didn't have the time to. Please leave it open until I add some tests as well :)

pavlospt:
@passy Ok added appropriate test cases :) It's ready for review :)

PR #379, opened 2018-06-01
Port `animations` to `sample-kotlin`

pavlospt:
This PR follows a similar path as the previous one :) It ports `animations` package from Java `sample` to Kotlin `sample-kotlin`.

A nice side effect of it is that even more parts of the framework are being "tested" in Kotlin code, which will allow catching Kotlin-related issues earlier.

### DIFF
--- a/sample-kotlin/src/main/java/com/fblitho/lithoktsample/NavigatableDemoActivity.kt
+++ b/sample-kotlin/src/main/java/com/fblitho/lithoktsample/NavigatableDemoActivity.kt
@@ -20,8 +20,9 @@ import android.support.v7.app.AppCompatActivity
 import android.view.MenuItem
 import com.fblitho.lithoktsample.demo.DataModels
 import com.fblitho.lithoktsample.demo.DemoListActivity
-import com.fblitho.lithoktsample.demo.DemoListActivity.Companion.DEFAULT_INDEX
-import com.fblitho.lithoktsample.demo.DemoListActivity.Companion.INDEX
+import com.fblitho.lithoktsample.demo.DemoListActivity.Companion.INDICES
+import com.fblitho.lithoktsample.demo.DemoListDataModel
+import java.util.Arrays
 
 @SuppressLint("Registered")
 open class NavigatableDemoActivity : AppCompatActivity() {
@@ -30,8 +31,8 @@ open class NavigatableDemoActivity : AppCompatActivity() {
     super.onCreate(savedInstanceState)
 
     intent
-        .getIntExtra(INDEX, DEFAULT_INDEX)
-        .takeIf { it != DEFAULT_INDEX }
+        .getIntArrayExtra(INDICES)
+        .takeIf { it != null }
         ?.let {
           supportActionBar?.setDisplayHomeAsUpEnabled(true)
           setTitleFromIndex(it)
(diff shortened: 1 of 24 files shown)


sample-kotlin/src/main/java/com/fblitho/lithoktsample/animations/animatedbadge/AnimatedBadgeSpec.kt
```diff
@@ -0,0 +1,191 @@
...
+                                  .heightDip(18f)
+                                  .widthDip(18f)
+                                  .background(buildRoundedRect(c, -0x48b5, 9)))
+                          .child(
+                              if (!expanded)
+                                null
+                              else
+                                Text.create(c)
+                                    .transitionKey(
+                                        TRANSITION_KEY_TEXT) // still need transition keys for
```
passy: Would be nice to have some more consistent formatting for the comments here.

sample-kotlin/src/main/java/com/fblitho/lithoktsample/animations/animationcomposition/ComposedAnimationsComponentSpec.kt
```diff
@@ -0,0 +1,75 @@
...
+
+@LayoutSpec
+object ComposedAnimationsComponentSpec {
+
+  @OnCreateLayout
+  fun onCreateLayout(c: ComponentContext): Component =
+      RecyclerCollectionComponent.create(c)
+          .disablePTR(true)
+          .section(
+              DataDiffSection.create<Any>(SectionContext(c))
```
passy: Is that `<Any>` required?

sample-kotlin/src/main/java/com/fblitho/lithoktsample/animations/animationcomposition/ComposedAnimationsComponentSpec.kt
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  @OnEvent(OnCheckIsSameItemEvent::class)
+  fun isSameItem(
+      c: ComponentContext,
+      @FromEvent previousItem: Data,
+      @FromEvent nextItem: Data
+  ): Boolean = previousItem.number == nextItem.number
+
+  private fun generateData(number: Int): List<Any> {
```
passy: Can we make this of type `<Data>`?

sample-kotlin/src/main/java/com/fblitho/lithoktsample/NavigatableDemoActivity.kt
```diff
@@ -30,29 +31,40 @@ open class NavigatableDemoActivity : AppCompatActivity() {
     super.onCreate(savedInstanceState)
 
     intent
-        .getIntExtra(INDEX, DEFAULT_INDEX)
-        .takeIf { it != DEFAULT_INDEX }
+        .getIntArrayExtra(INDICES)
+        .takeIf { it != null }
```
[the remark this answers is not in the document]
pavlospt: `.filterNotNull` would allow not null `Int`s to go through. In this context, we need `takeIf` because we only need to check the array if it is not null in its entirety!

sample-kotlin/src/main/java/com/fblitho/lithoktsample/animations/animationcomposition/ComposedAnimationsComponentSpec.kt
pavlospt: Unfortunately yes, because otherwise, it can not infer the generic type. In that case, `Any` is being replaced with `Data` in addition to the change of the following comment.

sample-kotlin/src/main/java/com/fblitho/lithoktsample/animations/animationcomposition/ComposedAnimationsComponentSpec.kt
passy: Oh, please don't ever feel just because we use some terrible antipattern in our code that you have to keep it around. :P

sample-kotlin/src/main/java/com/fblitho/lithoktsample/NavigatableDemoActivity.kt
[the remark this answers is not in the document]
passy: Ah, gotcha! Thanks for clarifying.

PR #382, opened 2018-06-06
Optimize Workflow for performance

pavlospt:
For some reason I had used the wrong Gradle command to pre-fetch dependencies (e.g.: `./gradlew dependencies`) and when reaching the `gradle_tests_run` Job in the Workflow, Gradle was downloading those dependencies every time.

I copied the Gradle command used in @passy 's Dockerfile of the previously used image, in order to trigger a dependencies fetch.


passy:
Amazing! Thanks for this.

PR #383, opened 2018-06-08
Allow EventData to return current instance

pavlospt:
* Using `apply` function from Kotlin, we can have a Builder pattern-like behavior, allowing call chaining.


pavlospt:
You mean `..) : EventData = apply { ... ` ?

passy:
Sorry for being so slow. Yes. :)

pavlospt:
@passy Great, will add it! Also just out of curiosity, why would you want this?

passy:
@pavlospt I like type signatures. :) I think it's good practice to keep them around unless you're returning `Unit`.

Issue #381, opened 2018-06-05
Border edge values work on opposite edges of those wanted


pavlospt:
- [x] I have searched [existing issues](https://github.com/facebook/litho/issues) and this is not a duplicate

## Version
Latest version of Litho, I think it is 0.15.0.

## Issues and Steps to Reproduce
Border component does not seem to render path effect on the requested Edges.

Create a `Border` component like following and specify a path Effect on `VERTICAL` Edges:

```kotlin
Border.create(c)
                  .color(YogaEdge.VERTICAL, NiceColor.RED)
                  .widthDip(YogaEdge.ALL, 5)
                  .dashEffect(floatArrayOf(20f, 5f), 0f)
                  .build())
```

## Expected Behavior
Expected behaviour is to draw path effect on Vertical edges Left and Right.

## Link to Code
https://github.com/facebook/litho/blob/master/litho-core/src/main/java/com/facebook/litho/Border.java#L128 

The above method is responsible to set the edges, but seems like it is setting values on an opposite fashion of what the requested Edge was: 

```java
case VERTICAL:
        edges[EDGE_TOP] = value;
        edges[EDGE_BOTTOM] = value;
        break;
case HORIZONTAL:
        edges[EDGE_LEFT] = value;
        edges[EDGE_RIGHT] = value;
        break;
```

### Screenshot from `sample`
![screenshot_20180605-115334__01](https://user-images.githubusercontent.com/1070871/40966320-dda80582-68b8-11e8-9e2b-95ab82c937e9.jpg)

I can prepare a PR fixing this, in case it is indeed wrong and not a wrong behaviour!

xiphirx:
It seems like it is correct currently, based off Yoga's source: https://github.com/facebook/yoga/blob/master/yoga/Yoga.cpp#L125

pavlospt:
@xiphirx Is there a chance that it might be the opposite there as well? It feels weird to write: `.color(YogaEdge.VERTICAL, NiceColor.RED)` and have the path be drawn on bottom and top, or this is just me 😂

marco-cova:
Yep @xiphirx is right... as from Yoga website documentation: 
> BORDER in Yoga acts exactly like padding 
In padding, "Vertical" is applied to Top/Bottom edges. I honestly don't know if Border should be reversed... but if so, it should be done within Yoga.

pavlospt:
Thanks for explaining @marco-cova ! it totally makes sense, If it is a domain specific variation. I will close the PR and this issue then :)

PR #384, opened 2018-06-11
Fix Border opposite orientation edges

pavlospt:
* In Border class setting a path effect on edges, was using the opposite edges of those requested in order to apply the effect.

* Addresses #381


xiphirx:
Actually looking at Yoga's source code, it does seem like vertical == top/bottom and horizontal == left/right

https://github.com/facebook/yoga/blob/master/yoga/Yoga.cpp#L125

PR #393, opened 2018-06-29
Update kotlin-stdlib-jre7 to kotlin-stdlib-jdk7

pavlospt:
* Kotlin JRE7 dependency has been deprecated and replaced with JDK7 version.


passy:
Thanks!

PR #394, opened 2018-06-29
Return Map instead of MutableMap on SampleComponentsLogger

pavlospt:
* Probably this was caused from a Java to Kotlin conversion. Since Kotlin
also has Map we could return an immutable Map instead of a Mutable :)


passy:
Good catch. Thanks!

PR #402, opened 2018-07-21
Fixed code example in mount-specs.md

teganburns:
The code example ( https://fblitho.com/docs/mount-specs ) may confuse readers. Undefined variable "width" should have been "size.width" and cast as an int.


IanChilds:
Hi @teganburns, thanks for doing this! `size.width` in Litho is actually already an int, so we don't need the cast.

PR #409, opened 2018-08-01
Sections tutorial sample code didn't compile

oprisnik:
The DataDiffSection requires a type to be supplied when it's created, see https://fblitho.com/javadoc/com/facebook/litho/sections/common/DataDiffSection.html#create-com.facebook.litho.sections.SectionContext-


passy:
Good catch!

PR #431, opened 2018-09-18
`View`s  changed to `Views`

alikhaaki:


passy:
Thanks, but this is intended. `View` is an Android class. The plural doesn't exist in code, so we're referring to it as `View`s.

passy:
Thanks, but this is intended. `View` is an Android class. The plural doesn't exist in code, so we're referring to it as `View`s.

PR #448, opened 2018-10-21
Update Yoga SNAPSHOT version

pavlospt:
Yoga has only released a 1.10.0-SNAPSHOT version and not a 1.10.1, which breaks the `Publish snapshot` job of the CircleCI workflow.

https://oss.sonatype.org/content/repositories/snapshots/com/facebook/yoga/yoga/

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -92,7 +92,7 @@ ext.deps = [
 // For releases, we want to depend on a stable version of Yoga.
 ext.deps.yoga =
     isSnapshot()
-        ? 'com.facebook.yoga:yoga:1.10.1-SNAPSHOT'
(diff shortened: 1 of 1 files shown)


muraziz:
Oh, my bad, thanks @pavlospt for fixing this!

muraziz:
Actually @passy is sending another fix, changing snapshot version of yoga instead of its reference in litho. This will be fixed by his diff.

pavlospt:
@muraziz ok this makes sense!

muraziz:
This commit in yoga should fix it!
https://github.com/facebook/yoga/commit/e12b78f99bf7d0d46a74da48f5a22cc849f8cd73

PR #449, opened 2018-10-23
Upgrade Gradle plugin

passy:
Summary:
This brings a new build tools requirements and NDK with it.
The `gnustl_shared` version is also no longer supported, but that
only matters for debug builds anyway.

Test Plan:
```
./gradlew :sample:installDebug
```

Also, let's see first if Circle is okay with this.


passy:
Circle CI hasn't updated their images yet, so this isn't really worth it.

PR #455, opened 2018-10-31
if has ImageSpan,  getTextWidths return wrong width. So change to get…

keito5656:
If Text has `ImageSpan`,  `getTextWidths` return wrong width. So change to `getLineWidth`.


astreet:
Can you provide a test case where this doesn't work properly? Changing this to use the line width won't be correct unfortunately :(

astreet:
Closing to clean up -- re-open if you would like to continue on this PR :)

PR #464, opened 2018-11-27
Bump Flipper to v0.12

passy:
Summary:
Using the most recent version.

Test Plan:
`./gradlew :sample:installDebug` completes.
Kinda needed a test surface as we've had some trouble publishing the artifacts ...


pavlospt:
@passy shouldn't a script updating the dependency on Litho and other OS Facebook projects using Flipper, be enough to verify the publish of any new version of Flipper?

pavlospt:
@passy Bintray has Webhooks. Ideally there would be a receiver for Flipper's publish webhook that would open that PR https://bintray.com/docs/api/#_register_a_webhook

passy:
@pavlospt Yes, definitely possible to build, but unlikely I'll find time for that before long ...

PR #478, opened 2019-01-18
Upgrade gradle 511

pavlospt:
* Bump Gradle version to 5.1.1 stable
* Bump Kotin version to 1.3.11
* Update test resources copy task for Gradle 5.0+ 

### Run without the workaround
image

### Run with the updated workaround
image


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,21 +1,5 @@
-#
```
passy: Can you restore the header please? :)

litho-it/build.gradle
```diff
@@ -89,30 +89,24 @@ dependencies {
...
+    // Base path which is recognized by android studio.
+    def testClassesPath = "${buildDir}/intermediates/classes/test/"
+    // Copy must be done for each variant.
+    def variants = android.libraryVariants.collect()
+
+    variants.each { variant ->
+        def variationName = variant.name.capitalize()
+        def variationPath = variant.buildType.name
+
+        task "copyTest${variationName}Resources"(type: Copy) {
```
passy: Ahh, that's how you do this now. Thanks for looking into this.

pavlospt: @passy `Copy` was always there! It's just that now you need to work the task execution through Task dependencies, or utilize Worker API through a custom plugin that would host the task (which would be overkill in that case IMO). Since we need the `test` task to always run after those tasks have executed, we just add a dependency on it, for the created tasks.

passy: I didn't realise you could just create an ad-hoc task like this and depend on it. If I remember correctly, they removed the `.execute()` method on tasks you create imperatively, didn't they?

pavlospt: Yeap, it's removed and the project does not build with it. This Groovy syntax has the same effect as old .create() but registers it. In KotlinDSL we would explicitly register it and then configure it!

Issue #482, opened 2019-01-29
Support for incremental annotation processing in Gradle


harliedharma:
- [x] I have searched [existing issues](https://github.com/facebook/litho/issues) and this is not a duplicate

Gradle supports incremental annotation processing since version 4.7 to speed up compilation process. 
For details see:
https://docs.gradle.org/nightly/userguide/java_plugin.html#sec:incremental_annotation_processing

https://github.com/gradle/gradle/issues/5277

Litho currently doesn't support incremental annotation processing. AFAIK, by default, it will make Gradle disable incremental annotation processing for the entire module. It will be great if Litho enable this feature as Litho heavily depends on annotation processing.

pavlospt:
I can take a look at this and give it a try!

pavlospt:
@harliedharma the PR for incremental annotation processing has been merged! You can close this :)

karandeep-swiggy:
I am on the latest version(0.33.0) and I am still getting this error:
`Incremental annotation processing requested, but support is disabled because the following processors are not incremental: com.facebook.litho.specmodels.processor.testing.ComponentsTestingProcessor (NON_INCREMENTAL)`

PR #483, opened 2019-01-29
[Gradle] Enable incremental AnnotationProcessing

pavlospt:
This PR handles required changes to make Litho compatible with Gradle incremental annotation processing. 

## Changes needed
* Add originating element when creating a `*SpecModel`.
    * This is required, in order to pass it on the `TypeSpec` that in turn forwards those to `Filer API`.
* Add relevant Gradle META-INF file that specify which annotation processors support incremental AP and what is their type. (I explicitly left out any `*TestingProcessor` in order to avoid any false alarms when running tests) 

## Results
I did some manual tests and it seems that there is indeed a performance improvement between this branch and `master`, but I am not quite sure if I am biased or this is working in reality. I have some screenshots indicating that less tasks were executed for the same changes between `pavlospt:master` (rebased on latest `litho:master`) and `pavlospt:incremental_annotation_processing`. 

### Java changes on `master`
Java changes on `master`

### Java changes on `incremental_annotation_processing`
Java changes on `incremental_annotation_processing`

## Test plan 
I am not sure if this can be tested. Other implementations of IAP i have found around, are making use of a `Map` inside the `Processor` and return the `orginatingElement` from an equivalent method there. Essentially they test the existence of an element's FQCN in the `Map`. 

The way I implemented the current solution, is just by taking advantage of `TypeSpec` 's ability to hold an `originatingElement` which then passes it to the `Filer API`. I guess we can trust Square's tests for this 😄 

## Related Issue
Addresses issue #482


litho-processor/src/main/java/com/facebook/litho/specmodels/processor/MountSpecModelFactory.java
```diff
@@ -109,6 +109,7 @@ public MountSpecModel create(
       @Nullable DependencyInjectionHelper dependencyInjectionHelper,
       @Nullable InterStageStore interStageStore) {
     return new MountSpecModel(
+        element,
```
passy: I wonder if it's still useful to pass in the other arguments that are directly derived from `element`.

litho-testing/src/main/java/com/facebook/litho/testing/specmodels/MockSpecModel.java
```diff
@@ -202,6 +203,11 @@ private MockSpecModel(
     mEnclosedSpecModel = enclosedSpecModel;
   }
 
+  @Override
+  public TypeElement getOriginatingElement() {
+    return null;
```
passy: Can this be null under normal (production) circumstances? If not, is there a way to create a no-op `TypeElement` instead to make sure we keep this non-nullable throughout?

litho-processor/src/main/java/com/facebook/litho/specmodels/processor/MountSpecModelFactory.java
pavlospt: @passy I was wondering exactly the same thing! We could also then create separate `ElementUtils` that would contain info extraction methods, replacing essentially all of those assignments in `*SpecModel` constructors. This would though imply a bigger refactor! If this is something that the Litho team would feel ok with, I would be more than happy to do it. 

I believe thought that this would only apply to `*SpecModel` implementing classes and not `TestSpecModel` or `MockSpecModel` which do not contain an originating element in purpose.

litho-testing/src/main/java/com/facebook/litho/testing/specmodels/MockSpecModel.java
pavlospt: @passy Normally it should never be `null`! I can take a look to check if I can create a no-op `TypeElement`! I believe that this would be a good solution since there is no `TestingProcessor` registered for incremental annotation processing, thus avoiding any possible issues in tests.

pavlospt: @passy Added a `NoOpTypeElement` and removed `@Nullable` annotation from `SpecModel`. Tests were green locally. Waiting for CI as well.

passy:
And feel free to ignore the buck issues. I can look into that. :)

pavlospt:
@passy Where is this referring to? 
> This won't work, as you don't know that the underlying element is a TypeElement. However, you can use getRepresentedObject and cast it

passy:
Ideally @IanChilds could chime in here directly. :)

pavlospt:
Sorry for pinging this PR again, but is there any chance we can get Ian's explanation on that comment, so that I could update it accordingly :) ?

IanChilds:
Sorry, I missed this. The problem is that there is no guarantee that the underlying element was a TypeElement (for example you could construct a SpecModel from a PsiClass).

Instead, you could use getRepresentedObject, which is the underlying element (your name is better), and then cast that to be a TypeElement in your case, where you know that it is a TypeElement.

pavlospt:
@IanChilds Hmm ok will update it and push again :)

pavlospt:
@IanChilds I updated the implementation to use `.getRepresentedObject()` from `SpecModel`, but I kept the `getOriginatingElement()` API as well, in order to keep a clear distinction between the usage of those two!

IanChilds:
I'm afraid this is still wrong. We shouldn't have a reference to `TypeElement` inside the SpecModel. Instead, wherever you are calling `getOriginatingElement` outside of the SpecModel, just call `getRepresentedObject()` and do the cast there.

pavlospt:
@IanChilds Ok will remove that one then as well! Thought you were positive on keeping that method, my bad :)

pavlospt:
@IanChilds I updated again! I hope this time it will be as you wanted :D 🙏

pavlospt:
@facebook-github-bot yeap it is still relevant! :)

litho-processor/src/main/java/com/facebook/litho/specmodels/model/DefaultLayoutSpecGenerator.java
```diff
@@ -46,6 +47,7 @@ public TypeSpec generate(LayoutSpecModel layoutSpecModel, EnumSet<RunMode> runMo
     final TypeSpec.Builder typeSpec =
         TypeSpec.classBuilder(layoutSpecModel.getComponentName())
             .superclass(ClassNames.COMPONENT)
+            .addOriginatingElement((TypeElement) layoutSpecModel.getRepresentedObject())
```
passy: Is this actually being used somewhere now? And if not, are we still getting the speed-up benefits from what it is now?

pavlospt: The following is still true, despite reverting the changes to the `SpecModel`:
> Add originating element when creating a `*SpecModel`.
> * This is required, in order to pass it on the `TypeSpec` that in turn forwards those to `Filer API`.

IanChilds: Where is the code with the Filer API? I'm not seeing much code remaining after the latest update :)

pavlospt: You can check this test from Javapoet, which does exactly this assertion https://github.com/square/javapoet/blob/master/src/test/java/com/squareup/javapoet/FileWritingTest.java 

Also please check `Filer` docs: https://docs.oracle.com/javase/7/docs/api/javax/annotation/processing/Filer.html 

> If there are no originating elements, none need to be passed. This information may be used in an incremental environment to determine the need to rerun processors or remove generated files. Non-incremental environments may ignore the originating element information.

IanChilds: Oh my mistake, I read this totally wrong! Thanks for your patience :) I leave this to @passy

pavlospt: No problem! Happy to explain!

pavlospt:
@passy Hey Pascal, sorry for pinging again! Is this ok to be merged or is still being reviewed internally?

passy:
Sorry, Pavlos. I'm stretched a bit thin at the moment. I'll have look where we left this off.

pavlospt:
Sure thing @passy ! I was just wondering if there is any more work to be done :)

litho-processor/src/main/java/com/facebook/litho/specmodels/model/DefaultLayoutSpecGenerator.java
passy: @IanChilds raised to me that there is a possibility of the represented object being a `PsiClass`. Could you add a check here before casting to ensure that we don't blow up in that case? And sorry for the churn on this.

litho-processor/src/main/java/com/facebook/litho/specmodels/model/DefaultLayoutSpecGenerator.java
```diff
@@ -47,9 +47,12 @@ public TypeSpec generate(LayoutSpecModel layoutSpecModel, EnumSet<RunMode> runMo
     final TypeSpec.Builder typeSpec =
         TypeSpec.classBuilder(layoutSpecModel.getComponentName())
             .superclass(ClassNames.COMPONENT)
-            .addOriginatingElement((TypeElement) layoutSpecModel.getRepresentedObject())
             .addTypeVariables(layoutSpecModel.getTypeVariables());
 
+    if (SpecModelUtils.isTypeElement(layoutSpecModel)) {
```
pavlospt: @passy @IanChilds Not sure if this is something that adheres to Litho's code style. If not I could extract that to a `boolean` and use that in the `if` condition. 
I also thought about creating a `TypeSpecHolder` generator responsible to add the `OriginatingElement`, but that would seem "overkill" in that case. If that's a better approach in your opinion though, let me know and I good get something up.

passy: It looks good to me. Thanks putting so much thought in it. :)

PR #484, opened 2019-01-29
Address a leak of LayoutState

tnorman42:
This is due to MountState not releasing its reference to LayoutState even after unmounting all items.

### DIFF
--- a/litho-core/src/main/java/com/facebook/litho/LayoutState.java
+++ b/litho-core/src/main/java/com/facebook/litho/LayoutState.java
@@ -2048,6 +2048,11 @@ class LayoutState {
     return mPreviousLayoutStateId;
   }
 
+  @VisibleForTesting
+  int getReferenceCount() {
+    return mReferenceCount.get();
+  }
+
   /**
    * See {@link LayoutState#acquireRef} Call this when you are done using the reference to the
    * LayoutState.
--- a/litho-core/src/main/java/com/facebook/litho/LithoView.java
+++ b/litho-core/src/main/java/com/facebook/litho/LithoView.java
@@ -787,6 +787,9 @@ public class LithoView extends ComponentHost {
       mComponentTree = null;
       mNullComponentCause = "release_CT";
     }
+
+    mMountState.unmountAllItems();
+    mMountState.releaseLastMountedLayoutState();
   }
 
   void mount(LayoutState layoutState, Rect currentVisibleArea, boolean processVisibilityOutputs) {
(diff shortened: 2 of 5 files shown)


astreet:
Thanks for the PR! This seems more appropriate in `LithoView.release` since we don't know *why* we're unmounting all items here -- what is the case in which you're running into this?

tnorman42:
Ah yes, good point. But LithoView.release() is not always called in cases where MountState should be releasing its LayoutState reference. See, for example: RecyclerBinder.onViewRecycled. In this case, if the LithoView mounts something new, it will eventually call releaseRef() on the old LayoutState, but if the LithoView is discarded (e.g. RecyclerView goes away), then the LayoutState gets leaked.

Another place that would maybe make sense is when calling setComponentTree(null), but this is currently guarded by ComponentsConfiguration.unmountAllWhenComponentTreeSetToNull, which is default false. Why wouldn't this be true? If this config were true, then we could release the LayoutState from mMountState there.

I will update the PR to do this on release(), but RB will need to also call release() after setComponentTree(null), which is a bit odd because it will be reusing the LithoView after releasing it. What do you think, is there a better way?

This leak is becoming a problem for us because we are trying to add a Component lifecycle method for notifying a Component when it is no longer attached to a layout. This is important for us because one of our Components needs to dispose of an object at this time.

PS I'm collaborating with vinc3m1

tnorman42:
Added mLastMountedComponentTreeId = ComponentTree.INVALID_ID - which handles the case where RB reuses the LithoView for the same ComponentTree.

Any feedback on this PR?

muraziz:
Hey @tnorman42 , thanks for PR :) Would you mind rebasing to master and updating commit? We recently got rid of pooling of internal objects which is conflicting with this commit. This particular commit (https://github.com/facebook/litho/commit/9327eafef95bfcbce3c0517ef14ed8c79d5b199e) is conflicting with this one.

tnorman42:
@muraziz yikes, well this basically makes our approach impossible. Let me describe what we need, and maybe you have another idea how we can do it.

We have a Component that needs to dispose of an expensive resource when it's not attached to a Layout. Our approach was going to be to use the LayoutState refcount and add refcounting for Components interested in this lifecycle method. However, with your latest change, there's no more LayoutState refcounting.

@State won't work in this case because State is too long-lived and RecyclerBinder holds onto State for the entire feed. (Plus there's no lifecycle callbacks for State disposal anyway.)

We don't want to have to rely on the garbage collector for this.

muraziz:
@tnorman42 our recent change was about no longer using pools for internal objects including `LayoutState`. 

> We have a Component that needs to dispose of an expensive resource when it's not attached to a Layout. Our approach was going to be to use the LayoutState refcount and add refcounting for Components interested in this lifecycle method. However, with your latest change, there's no more LayoutState refcounting.

Regarding the lifecycle of the expensive resource, should it be limited to the time when it is on viewport or just lifecycle of activity/fragment? If it is the latter, we are working on new API @OnDestroyState that will be called when that happens.

tnorman42:
Closing this PR as it is obsolete with the recent pooling and refcounting changes.

PR #489, opened 2019-02-13
Expose incremental mount option in RecyclerBinder configuration builder

yuntao:
This allows a RecyclerBinder to be built with the flexibility of toggling incremental mount for its children Litho views.

We have encountered "animated parent view" bugs, which needed this option as a fix. When a Litho view is used as a child inside an animated parent view, the Litho view does not render properly and could show up as blank when incremental mount is enabled, because incremental mount doesn't get triggered during animation and the parent view does not propagate setHasTransientState(boolean) calls to the contained Litho view. When this happens, disabling incremental mount on the Litho view is a fix.

This works until a RecyclerCollectionComponent is used in the Litho view. As there is no way to disable incremental mount in the RecyclerCollectionComponent's underlying RecyclerBinder, the blank rendering bug would still manifest for children inside the RecyclerBinder.

With this change, we can disable incremental mount using RecyclerBinderConfiguration:

```
RecyclerCollectionComponent.create(c)
    .recyclerConfiguration(
        ListRecyclerConfiguration.create()
            .recyclerBinderConfiguration(
                RecyclerBinderConfiguration.create()
                    .useIncrementalMount(true/false)
                    .build())
            .build())
    .build()
```

cc @vinc3m1

### DIFF
--- a/litho-it/src/test/java/com/facebook/litho/widget/RecyclerBinderTest.java
+++ b/litho-it/src/test/java/com/facebook/litho/widget/RecyclerBinderTest.java
@@ -159,6 +159,7 @@ public class RecyclerBinderTest {
               RenderInfo renderInfo,
               LayoutHandler layoutHandler,
               boolean useSharedLayoutStateFuture,
+              boolean useIncrementalMount,
               ComponentTreeMeasureListenerFactory componentTreeMeasureListenerFactory,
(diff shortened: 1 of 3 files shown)


colriot:
Fixed in v0.25.0, 241e850

PR #492, opened 2019-02-13
Replace usages of Powermock in litho-testing with local replacement

### LINKED ISSUE #490: Remove references to Powermock in litho-testing

We are using litho for our view tests, but we would like to remove our usages of Powermock. The `litho-testing` package that we use has several reference to 1 class in Powermock: `Whitebox`.

Since this is only 1 class, it is quite easy to have a local replacement with just the functions that `litho-testing` needs. [This gist](https://gist.github.com/TimvdLippe/1635f0f1606dca14d6c91d43ec30fac3) contains the required local copy of Whitebox that we already succesfully used internally. However, we would like to make this change upstream as well.

Would you consider remove these few references to Powermock in `litho-testing` with the local replacement implementation in `com.facebook.litho`?

TimvdLippe:
This allows us to remove the Powermock dependency from litho-testing.

Fixes #490

### DIFF
--- a/litho-testing/src/main/java/com/facebook/litho/testing/Whitebox.java
+++ b/litho-testing/src/main/java/com/facebook/litho/testing/Whitebox.java
@@ -1,0 +1,108 @@
+package com.facebook.litho.testing;
+
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+
+/** Internal copy of Whitebox implementation of Powermock. */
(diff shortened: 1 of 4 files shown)


vinc3m1:
This can be closed in favor of #493

TimvdLippe:
Ah I was not aware of the same usages in `litho-it`. Thanks for the followup @vinc3m1 !

PR #494, opened 2019-02-15
Add support for visibility handling when incremental mount is disabled

yuntao:
Visibility event processing is not support when incremental mount is disabled currently. This change adds approximate support based on the LithoView's layout, attach, and detach lifecycle invocations. 

- [ ] Add tests
- [ ] Update docs


litho-core/src/main/java/com/facebook/litho/LithoView.java
```diff
@@ -206,12 +206,25 @@ private void onAttach() {
...
+      if (!isIncrementalMountEnabled() && !isMountStateDirty() && !mountStateNeedsRemount()) {
+        dispatchVisibleEvent();
+      }
     }
   }
 
   private void onDetach() {
     if (mIsAttached) {
       mIsAttached = false;
+
```
pasqualeanatriello: I think we already call invisible in onDetach. Is this really needed?

PR #498, opened 2019-02-22
Add onScrollChangeListener property to HorizontalScrollSpec.

tnorman42:
This matches the property on VerticalScrollSpec.

### DIFF
--- a/litho-widget/src/main/java/com/facebook/litho/widget/HorizontalScrollSpec.java
+++ b/litho-widget/src/main/java/com/facebook/litho/widget/HorizontalScrollSpec.java
@@ -22,6 +22,7 @@ import static com.facebook.litho.SizeSpec.UNSPECIFIED;
 import android.content.Context;
 import android.content.res.TypedArray;
 import android.support.v4.util.Pools.SynchronizedPool;
+import android.view.View;
 import android.view.ViewTreeObserver;
 import android.widget.HorizontalScrollView;
 import com.facebook.litho.Component;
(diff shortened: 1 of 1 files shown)


litho-widget/src/main/java/com/facebook/litho/widget/HorizontalScrollSpec.java
```diff
@@ -203,6 +205,14 @@ public boolean onPreDraw() {
         new ViewTreeObserver.OnScrollChangedListener() {
           @Override
           public void onScrollChanged() {
+            if (onScrollChangeListener != null) {
+              onScrollChangeListener.onScrollChange(
```
muraziz: Our lint caught that `View.OnScrollChangeListener` is available only from API23 while we support API15+. Can you find alternative way of achieving this?

pavlospt: ..or API-version-guard it 😄 !

tnorman42: VerticalScrollSpec uses NestedScrollView.OnScrollChangeListener in the same way, and it's API24+. I don't see how it's being guarded there.

In my PR, we're only using the interface, so as long as Litho is built with SDK 23+, it should work. I'm not calling any SDK methods that needed to be guarded against.

muraziz: Actually `NestedScrollView` is implemented in support library and they provide their own OnScrollChangedListener, so that one is safe to use: 

https://android.googlesource.com/platform/frameworks/support/+/refs/heads/support-library-27.1.1/core-ui/src/main/java/android/support/v4/widget/NestedScrollView.java#84

muraziz: You could create that interface inside spec instead

astreet:
It looks like for some reason we're using a ViewTreeObserver here which is not correct -- we should get that fixed first and then we can add this listener :) @muraziz will send something out

PR #500, opened 2019-02-22
Add ability to specify a transition via Component.Builder.

tnorman42:
This is useful in cases where Components are created and/or decorated generically.

### DIFF
--- a/litho-core/src/main/java/com/facebook/litho/CommonPropsHolder.java
+++ b/litho-core/src/main/java/com/facebook/litho/CommonPropsHolder.java
@@ -456,6 +456,10 @@ class CommonPropsHolder implements CommonProps, CommonPropsCopyable {
     getOrCreateOtherProps().transitionKey(key);
   }
 
+  void transition(Transition transition) {
+    getOrCreateOtherProps().setTransition(transition);
+  }
+
   @Override
   @Nullable
   public String getTransitionKey() {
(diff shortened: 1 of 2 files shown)


muraziz:
@tnorman42 thanks for PR, can you elaborate more what you are trying to achieve here? This is quite divergent from our current model where transition and component declarations are separated.

tnorman42:
Sure, we create our Litho Components through a generic code path that uses a high-level description of the desired UI. We want to be able to support adding a Transition on any of these Components. It is much simpler to do this through the generic Component.Builder method than by having our dozen-plus ComponentSpecs all duplicate the same @OnCreateTransition code. Given that transitionKey is exposed by Component.Builder, it doesn't seem too much of a stretch to also allow transitions to be specified.

muraziz:
Regarding transitionKey that's to make sure we can associate ones in component declaration with the ones in transition declaration. I'm still not very sure about this change as it is breaking our initial mental model of having transitions separate from component declaration. Let me get feedback from other people as well

cc @sergeynv @astreet for more pair of eyes to look into this

pavlospt:
@passy I think that this bot might need some slight modifications on when it checks for outdated PRs 😂

pasqualeanatriello:
Sorry for the slow iteration on this. We are trying to understand the more broad implications this might have. Hopefully we'll have an update by the end of the week

PR #501, opened 2019-02-22
Override all methods that might cause an NPE accessing mYogaNode.

tnorman42:
We were seeing a crash here because we had some generic Component.Builder code that was setting things like aspectRatio on Components that could ultimately end up being an EmptyComponent.


muraziz:
cc @davidaurelio, @pasqualeanatriello

davidaurelio:
Looks reasonable to me, but I will leave it to the Litho folks. cc @pasqualeanatriello @astreet

astreet:
This will land but is blocked on some internal tests

PR #502, opened 2019-02-22
Add a onComponentTreeReleased method to RenderInfo.

tnorman42:
This allows for cleanup when RecyclerBinder disposes of the ComponentTree.


muraziz:
We are working towards some sort of ondestroy callback which could be overlapping with this PR. Let me pull in and ask for feedback from folks.

astreet:
Is this only as a workaround for the lack of onDestroy in the meantime?

tnorman42:
> Is this only as a workaround for the lack of onDestroy in the meantime?

Yes, if we had an onDestroy lifecycle method, we wouldn't need this. So if you'd rather wait until that kind of lifecycle method might be available, we can just keep this as a local change until then.

muraziz:
@tnorman42 yeap, let's do that, we'll be providing onDestroy callback pretty soon.

PR #506, opened 2019-03-02
Fix BUCK environment setup

passy:
Install missing build tools. Bringing the change from D14280910 forward.


muraziz:
It's been landed as part of https://github.com/facebook/litho/commit/2354054a904897a895ca982d98b0f03f3a537cc7

PR #510, opened 2019-03-05
Unwrap RippleDrawable before setting it as the background for MountSpec.

tnorman42:
RippleDrawable does not display properly when wrapped with ComparableDrawable.

There may be other cases where this is a problem, but this is the only one we've identified so far.

### DIFF
--- a/litho-core/src/main/java/com/facebook/litho/MountState.java
+++ b/litho-core/src/main/java/com/facebook/litho/MountState.java
@@ -61,6 +61,7 @@ import androidx.core.view.ViewCompat;
 import com.facebook.infer.annotation.ThreadConfined;
 import com.facebook.litho.animation.AnimatedProperties;
 import com.facebook.litho.config.ComponentsConfiguration;
+import com.facebook.litho.drawable.DefaultComparableDrawable;
 import com.facebook.litho.reference.Reference;
 import java.util.ArrayList;
 import java.util.BitSet;
@@ -2177,6 +2178,11 @@ class MountState implements TransitionManager.OnAnimationCompleteListener {
   private static void setBackgroundCompat(View view, Drawable drawable) {
     if (Build.VERSION.SDK_INT < Build.VERSION_CODES.JELLY_BEAN) {
       view.setBackgroundDrawable(drawable);
+    } else if (Build.VERSION.SDK_INT >= 21
+        && drawable instanceof DefaultComparableDrawable
+        && ((DefaultComparableDrawable) drawable).getWrappedDrawable() instanceof RippleDrawable) {
+        // RippleDrawable does not work properly when it is wrapped, so unwrap it when setting it.
+        view.setBackground(((DefaultComparableDrawable) drawable).getWrappedDrawable());
     } else {
       view.setBackground(drawable);
     }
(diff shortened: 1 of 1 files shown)


sergeynv:
@adityasharat can you take a look?

adityasharat:
@tnorman42 could you also mention what is it that doesn't display correctly (maybe add a screenshot or a clip). Maybe we can fix the exact issue instead of unwrapping the drawable.

In addition, IMO it's not a good idea unwrap the drawable. Consider the following usage:

```java
return SomeComponent.create(c)
             .background(new CustomRippleDrawable());
```

Here `CustomRippleDrawable` extends `ComparableDrawableWrapper`. Given the proposed changes we will be breaking the assertion that the developer holds about which drawable will be set on the view.

If we really want to unwrap drawables then we should unwrap it for all drawables not just `RippleDrawable` for the sake of consistency. Which I believe is also not possible; consider the following usage:

```java
return SomeComponent.create(c)
             .background(new MyDrawable());
```

Here `MyDrawable` extends `ComparableDrawable`, this drawable cannot be unwrapped (i.e. doesn't extend `ComparableDrawableWrapper`, which again breaks the developers expectations. 

I think we should try to fix the exact issue if that fails we can consider unwrapping conditionally. Second opinions?

cc: @astreet, @pasqualeanatriello

tnorman42:
To be honest, I don't think it's a good idea to be doing ANY wrapping of Drawables yourself to begin with. If you want to wrap, you should use Android's DrawableWrapper, and not your own hand-rolled one. Look at what the Android support library implementations of DrawableWrapper do, e.g. DrawableWrapperApi21.java. You will have to do ALL of that to have a legitimate wrapper that has a chance of working. Do you really want to put yourself in a position of continuing to maintain this forked code?

IMO, you should either:

1. Make your ComparableDrawable class a dumb wrapper class that doesn't try to be a Drawable. Then you always unwrap it before setting it on a View. This would be quite safe because you never pass the OS a faked-out wrapper and don't have to maintain all the complicated wrapper code that DrawableWrapper already has.

2. (less preferred) Make your ComparableDrawable extend DrawableWrapper instead of Drawable so you can get the support library's and Android's correct wrapper implementation. This could get to be weird for the few cases where you want to extend ComparableWrapper as its own Drawable.

adityasharat:
Agreed; I actually like the first idea, i.e. make `ComparableDrawable` just a container for a drawable and not extend `Drawable` at all. Let me get back to you on this.

colriot:
@tnorman42 some work is going on currently, following your discussion with @adityasharat. We plan to ship this refactoring at the beginning of the next month.

anushreeag:
@colriot  @adityasharat , Any updates on this PR.

adityasharat:
Oh my 😞 
Yeah, this one has taken far longer than I would have anticipated. This slipped under my radar, I will try and push this by next week.

anushreeag:
> Oh my 😞
> Yeah, this one has taken far longer than I would have anticipated. This slipped under my radar, I will try and push this by next week.

Also please check https://github.com/facebook/litho/pull/606 which is dependent on this PR. Thanks

adityasharat:
We are planning to convert `ComparableDrawable` into an interface; so no more wrapping required. So we may not require these changes

adityasharat:
with 2bc34a985a2b3f221ebf6bfb6cd8e448fd538d17 this should no longer be an issue. `ComparableDrawable` is now an optional interface; so non comparable drawables will not be wrapped. We can close this.

PR #526, opened 2019-04-02
Eliminate Gradle deprecated methods

pavlospt:
There were some warnings for deprecated stuff that will be removed in Gradle 6.0. This PR eliminates them and replaces the previous usages with the equivalent providers.

e.g.: `getArchiveClassifier()`, `javaCompileProvider`


passy:
Great stuff, as always. Thanks, @pavlospt!

PR #553, opened 2019-06-06
Use the injected treeProps for Layout PerfEvent

boduan402:
At the runtime, the ComponentContext's treeProps could be null. In this case, it would better to use the treeProps that provided by caller.


litho-core/src/main/java/com/facebook/litho/LogTreePopulator.java
```diff
@@ -38,6 +38,21 @@ private LogTreePopulator() {}
   @CheckReturnValue
   public static PerfEvent populatePerfEventFromLogger(
       ComponentContext c, ComponentsLogger logger, PerfEvent perfEvent) {
+    return populatePerfEventFromLogger(c, c.getTreeProps(), logger, perfEvent);
+  }
+
+  /**
+   * Annotate a log event with the log tag set in the context, and extract the treeprops from a
```
colriot: Seems that jdoc was just copy-pasted without modification

litho-core/src/main/java/com/facebook/litho/ComponentTree.java
```diff
@@ -1772,7 +1772,10 @@ private void calculateLayout(
     final PerfEvent layoutEvent =
         logger != null
             ? LogTreePopulator.populatePerfEventFromLogger(
-                mContext, logger, logger.newPerformanceEvent(mContext, EVENT_LAYOUT_CALCULATE))
+                mContext,
+                mContext.getTreeProps() == null ? treeProps : mContext.getTreeProps(),
```
colriot: According to the nearby usage, `context.getTreeProps` is used as a fallback. Let's utilize the same logic here: just passing `treeProps` in, and falling back in overridden `populatePerfEventFromLogger` method.

litho-core/src/main/java/com/facebook/litho/LogTreePopulator.java
boduan402: Thanks Sergey, updated.

litho-core/src/main/java/com/facebook/litho/ComponentTree.java
boduan402: Good point. updated.

boduan402:
I mess up the forked branch. Will create a new one.

PR #567, opened 2019-07-18
Allow Specs to define their own disappearTo transitions.

tnorman42:
Create AnimationBindings from the previous LayoutState to allow Specs to define their own disappearance transitions.

This may not be the best way to solve this. Open to feedback.

## Summary

If a Spec defines a disappearTo transition for itself, it will not animate (appearFrom works). This change causes TransitionManager to include transitions defined in the previous LayoutState to be included when creating animations, which allows those disappearTo animations to run.

## Changelog

A Spec can now define its own disappearTo transition.

## Test Plan

Tested manually with our own Specs. Let me know what other tests you'd like me to add.

### DIFF
--- a/litho-core/src/main/java/com/facebook/litho/TransitionManager.java
+++ b/litho-core/src/main/java/com/facebook/litho/TransitionManager.java
@@ -275,7 +275,29 @@ public class TransitionManager {
       }
     }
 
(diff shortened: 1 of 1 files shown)


passy:
Thanks for the contribution, @tnorman42! I don't have a lot of context on the transition APIs, but I'll find someone internally to review this.

PR #569, opened 2019-07-20
explicit type casting needed

pritishyuvraj:
Type casting is required otherwise there is an error message from the complier: ```error: incompatible types: List<integer> cannot be converted to List<Object>```


astreet:
Thank you!

PR #573, opened 2019-07-30
Add support for A11y headers in Litho.

anushreeag:
## Summary
Add support for **Accessibility Headers** in Litho.
This will help the users to choose to “Navigate by Headings” and ignore scrolling through each and every sub items under a heading. “Navigate based on Heading“ can be chosen from Local Context Menu with talkback on ( Gesture for local context menu - Swipe up then right).
[ViewCompat](https://developer.android.com/reference/kotlin/androidx/core/view/ViewCompat?hl=en#setaccessibilityheading) and [AccessibilityNodeInfoCompat](https://developer.android.com/reference/android/support/v4/view/accessibility/AccessibilityNodeInfoCompat#setheading) have required API’s which can indicate that a view or corresponding accessibilityNode is an “Accessibility Heading” when talkback is enabled.

## Changelog
Add support for **Accessibility Headers** in Litho.

## Test Plan
Verified locally by switching talkback on and navigate by headers. It works fine and no crash found.

### DIFF
--- a/litho-core/src/main/java/com/facebook/litho/Component.java
+++ b/litho-core/src/main/java/com/facebook/litho/Component.java
@@ -1014,6 +1014,11 @@ public abstract class Component extends ComponentLifecycle
       return getThis();
     }
 
+    public T accessibilityHeading(boolean isHeading) {
+      mComponent.getOrCreateCommonProps().accessibilityHeading(isHeading);
+      return getThis();
+    }
+
     /**
      * If true, component duplicates its drawable state (focused, pressed, etc.) from the direct
      * parent.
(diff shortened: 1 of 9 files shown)


anushreeag:
FYI @astreet @adityasharat . Please review.

litho-core/src/main/java/com/facebook/litho/Component.java
```diff
@@ -1014,6 +1014,11 @@ public T importantForAccessibility(int importantForAccessibility) {
       return getThis();
     }
 
+    public T accessibilityHeading(boolean isHeading) {
```
astreet: Can you add javadocs here?

blavalla:
This looks good to me from an accessibility perspective.  It's probably worth noting in the javadoc that this prop won't have any impact below API 19.

litho-core/src/main/java/com/facebook/litho/Component.java
anushreeag: Done.

anushreeag:
Added javadoc. Please review. Thanks

PR #574, opened 2019-08-01
Use the injected treeProps for Layout PerfEvent

boduan402:
At the runtime, the ComponentContext's treeProps could be null. In this case, it would better to use the treeProps that provided by caller.

### DIFF
--- a/litho-core/src/main/java/com/facebook/litho/LogTreePopulator.java
+++ b/litho-core/src/main/java/com/facebook/litho/LogTreePopulator.java
@@ -26,10 +26,26 @@ import javax.annotation.CheckReturnValue;
 public final class LogTreePopulator {
   private LogTreePopulator() {}
 
+    /**
+     * Annotate a log event with the log tag set in the context, and extract the treeprops from a
+     * given {@link ComponentContext} and convert them into perf event annotations using a {@link
+     * ComponentsLogger} implementation.
+     *
+     * @return Annotated perf event, or <code>null</code> if the resulting event isn't deemed worthy
+     *     of reporting.
+     */
+    @Nullable
+    @CheckReturnValue
+    public static PerfEvent populatePerfEventFromLogger(
+            ComponentContext c, ComponentsLogger logger, @Nullable PerfEvent perfEvent) {
+      return populatePerfEventFromLogger(c, c.getTreeProps(), logger, perfEvent);
+    }
+
   /**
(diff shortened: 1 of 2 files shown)


boduan402:
Hi @colriot , can you please help review this change again? I somehow mess up the old branch and have to create this new pull request to address your comment. Could you please help review when getting a chance? Appreciate it.

litho-core/src/main/java/com/facebook/litho/LogTreePopulator.java
```diff
@@ -26,18 +26,34 @@
...
+     * ComponentsLogger} implementation.
+     *
+     * @return Annotated perf event, or <code>null</code> if the resulting event isn't deemed worthy
+     *     of reporting.
+     */
+    @Nullable
+    @CheckReturnValue
+    public static PerfEvent populatePerfEventFromLogger(
+            ComponentContext c, ComponentsLogger logger, @Nullable PerfEvent perfEvent) {
+      return populatePerfEventFromLogger(c, c.getTreeProps(), logger, perfEvent);
```
colriot: I suggest just passing `null` here and logic of the main method will decide what to use as a fallback

litho-core/src/main/java/com/facebook/litho/LogTreePopulator.java
```diff
@@ -50,7 +66,7 @@ public static PerfEvent populatePerfEventFromLogger(
 
     perfEvent.markerAnnotate(FrameworkLogEvents.PARAM_LOG_TAG, logTag);
 
-    @Nullable final TreeProps treeProps = c.getTreeProps();
+    @Nullable final TreeProps treeProps = c.getTreeProps() == null ? savedTreeProps : c.getTreeProps();
```
colriot: My previous comment in the original PR was about making logic of passing `treeProps` similar to other lib components, like `ComponentContext`. Pass nullable `treeProps` as a parameter to this method, if they are `null`, fallback to `c.getTreeProps()`.

boduan402: Thanks, fixed.

litho-core/src/main/java/com/facebook/litho/LogTreePopulator.java
boduan402: Thanks, fixed.

PR #583, opened 2019-08-06
Propagate isParentIncrementalMountDisabled flag when creating Section…

yuntao:
…Context from ComponentContext

## Summary

Commit https://github.com/facebook/litho/commit/c88a6605d8e77378e0ac89b267dcb260daddc6f4 introduced ComponentContext.isParentIncrementalMountDisabled flag to "only enable incremental mount if parent incremental mount is enabled". 

However, this doesn't work for any SectionSpec, because when the SectionContext is created from a ComponentContext, it doesn't propagate the isParentIncrementalMountDisabled flag forward.

## Changelog

Propagate isParentIncrementalMountDisabled flag when creating SessionContext from ComponentContext

## Test Plan

Existing tests.

### DIFF
--- a/litho-core/src/main/java/com/facebook/litho/ComponentContext.java
+++ b/litho-core/src/main/java/com/facebook/litho/ComponentContext.java
@@ -508,7 +508,7 @@ public class ComponentContext {
     }
   }
 
-  boolean isParentIncrementalMountDisabled() {
+  public boolean isParentIncrementalMountDisabled() {
     return mIsParentIncrementalMountDisabled;
   }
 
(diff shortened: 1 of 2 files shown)


yuntao:
With the latest https://github.com/facebook/litho/commit/87ff99ff9ce4143ac744d9755e0191ea2f691ff1 this PR is no longer required.

astreet:
Hey yes, I was going to comment that: we are trying to cleanup Component/ComponentContext to consolidate logic into ComponentTree/LayoutState/MountState -- if you need to figure out whether the parent has Inc Mount enabled, you can get the (nullable) CT off the ComponentContext and check if it has it enabled :)

PR #586, opened 2019-08-14
Fix ellipsizing with styled custom ellipsis text.

tnorman42:
## Summary

Litho was not measuring styled ellipsis text properly. Will include some before/after screenshots and test cases below.

## Changelog

Fixed text truncation with styled custom ellipsis text.

## Test Plan

Will include screenshots and test cases below.


tnorman42:
Before:
before

After:
after

Test case data:

Test case 1:
      text {
        content: "Same size truncation text '.'. This is a very b very c very d very e very f very g very h very i very j very k very l very m very n very o very p very q very r very s very t very u long line."
        style_runs {
          font_size: 20
        }
      }
      max_lines: 2
      truncation_text {
        content: "."
        style_runs {
          font_size: 20
          start_index: 0
          length: 1
        }
      }


Test case 2:
      text {
        content: "Same size truncation text '...'. This is a very b very c very d very e very f very g very h very i very j very k very l very m very n very o very p very q very r very s very t very u long line."
        style_runs {
          font_size: 20
        }
      }
      max_lines: 2
      truncation_text {
        content: "..."
        style_runs {
          font_size: 20
          start_index: 0
          length: 3
        }
      }

Test case 3:
      text {
        content: "Same size truncation text 'ABCDEFG'. This is a very b very c very d very e very f very g very h very i very j very k very l very m very n very o very p very q very r very s very t very u long line."
        style_runs {
          font_size: 20
        }
      }
      max_lines: 2
      truncation_text {
        content: "ABCDEFG"
        style_runs {
          font_size: 20
          start_index: 0
          length: 7
        }
      }

colriot:
cc @astreet @xiphirx who worked on the last ellipsis PR

astreet:
We're seeing some differences in our screenshot tests where it looks like it's now truncating more, but it's not clear to me why:

Before:
Screen Shot 2019-08-16 at 2 52 34 PM

After:
Screen Shot 2019-08-16 at 2 52 41 PM

There are a number of changes like this that don't seem like they should be happening. Any thoughts?

tnorman42:
@astreet I got some screenshots from you where this doesn't look correct. Can you provide the test data including all CharSequence spans to reproduce that test case? Thanks! We are tracking down one final truncation bug relating to the default font size that may be related.

astreet:
It's very hard for me to figure out exactly what code is creating the screenshot test that's failing (in relation to a normal Text.create()), let me see if Hilal knows a good way

tnorman42:
We were able to reproduce the problem. It seems to be connected to the
previous fix we sent. Will update this PR when we have a better solution.

tnorman42:
We have a fix ready internally that passes all of our tests including a testcase similar to the one you brought up. However, the fix is dependent on this other PR: https://github.com/facebook/litho/pull/535 So we will have to wait until that other PR lands to integrate this PR.

tnorman42:
I've updated this PR with our latest implementation that should more correctly handle all combinations of RTL languages and layout directions. Let me know if there are any other suspicious failures in your screenshot tests. Many thanks!

litho-widget/src/main/java/com/facebook/litho/widget/TextSpec.java
```diff
@@ -321,6 +321,12 @@ static void onMeasure(
       return;
     }
 
+    YogaDirection layoutDirection = layout.getResolvedLayoutDirection();
```
astreet: Can we leave this code in createTextLayout and keep the signature the same? It looks like we repeat this same pre-amble when we use it. I also am not a fan that you have to know to pre-process the alignment before passing it in (whereas now you mostly just pass everything in and it figures it all out)

tnorman42: The reason for pulling out the calls was to avoid performing duplicate calculations computing finalLayoutAlignment. We could address your concerns one of 3 ways: 1) keep the same signature for createTextLayout and just deal with the extra calculations, 2) Update createTextLayout with an in/out parameter for the finalLayoutAlignment to cache the result, or 3) extract those 3 lines to calculate finalLayoutAlignment into its own function but leave the rest of the code as is in the PR. Which approach would you prefer?

chggr:
This PR has been substituted by https://github.com/facebook/litho/pull/769

PR #593, opened 2019-09-10
Context -> ComponentContext

avikam:
error: Parameter in position 0 of a method annotated with interface com.facebook.litho.annotations.OnPrepare should be of type com.facebook.litho.ComponentContext.


avikam:
already proposed...

Issue #341, opened 2018-04-12
RippleDrawable 


mehcode:
- [x] I have searched [existing issues](https://github.com/facebook/litho/issues) and this is not a duplicate

## Version
```
litho_version = '0.13.1'
```
```gradle
implementation "com.facebook.litho:litho-core:$litho_version"
implementation "com.facebook.litho:litho-widget:$litho_version"
compileOnly "com.facebook.litho:litho-annotations:$litho_version"

kapt "com.facebook.litho:litho-annotations:$litho_version"
kapt "com.facebook.litho:litho-processor:$litho_version"
```

## Issues and Steps to Reproduce
 - Set an unmasked `RippleDrawable` as the background of a clickable component. 
 
## Expected Behavior

![native-view](https://user-images.githubusercontent.com/753919/38703025-002ccd68-3e57-11e8-95cf-53ae6d8398de.png) 

## Actual Behavior

![litho-view](https://user-images.githubusercontent.com/753919/38703035-06387266-3e57-11e8-937e-42e2863177f0.png)

## Link to Code
```kotlin
Row.create(c)
    .heightDip(50f)
    .child(
        Image.create(c)
            .widthDip(30f)
            .heightDip(30f)
            .marginDip(YogaEdge.LEFT, 15f)
            .marginDip(YogaEdge.TOP, 15f)
            .clickHandler(/* ... */)
            .backgroundRes(R.drawable.ripple)
            .drawableRes(R.drawable.ic_add))
```

*ripple.xml*
```xml
<?xml version="1.0" encoding="utf-8"?>
<ripple
    xmlns:android="http://schemas.android.com/apk/res/android"
    android:color="?android:colorControlHighlight">
</ripple>
```

muraziz:
TLDR; Unbounded ripple works only with `MountSpec`s

Basically the ability to draw outside the view bounds is kinda baked into the views. Drawing outside view bounds is possible only when view has background drawable AND drawable's `isProjected()` method is `true`. You can see that line here: http://androidxref.com/7.0.0_r1/xref/frameworks/base/core/java/android/view/View.java#13574. When view is invalidated it checks those two conditions and calls `damageInParent()` which allows to draw outside the bounds. 

[...]

PR #601, opened 2019-10-04
Set the RippleDrawable as a background of ComponentHost 

anushreeag:
Set the RippleDrawable as a background of ComponentHost  instead of mounting the RippleDrawable on ComponentHost.

Ripple Drawable are special since ripple should be projected outside the bounds of the view when set as a background.

This is a fix for setting the background only if its a Ripple Drawable.
For any other drawables which are set as background of a Component, we leave Litho to mount the non-RippleDrawables

## Summary
The issue which I am fixing here is same as reported in https://github.com/facebook/litho/issues/341.
Ripple Drawable gets clipped when set as a background for Row or Column components.

## Changelog

Fix Ripple effects not to be clipped when set as background for Row or Column

## Test Plan
Added enough unit tests and ripple works now. It gets projected outside the ComponentHost.


anushreeag:
@astreet @muraziz : FYI. Please review.

litho-core/src/main/java/com/facebook/litho/DrawableComponent.java
```diff
@@ -107,6 +108,26 @@ private void setDrawableWidth(int drawableWidth) {
...
+  /**
+   * Returns true if this component is a background for a ComponentHost or else returns false.
+   *
+   * <p>This will help to check if the background is a ripple drawable. We have to set
+   * RippleDrawables to the view's background so ripple can be projected outside the view bounds.
+   */
+  public boolean isBackground() {
+    return isBackground;
+  }
+
```
muraziz: How about moving most of Ripple Background checks here? In such way you don't need to do checks twice (during mount and during unmount) There could be two methods like
```
public void maybeSetIsRippleBackground() {
   if (SDK_INT < VERSION_CODES.LOLLIPOP) {
      return;
    }

    if (!(mDrawable instanceof DefaultComparableDrawable)) {
      return;
    }
    // We need to unwrap the RippleDrawable from DefaultComparableDrawable since
    // DefaultComparableDrawable does not override all the methods necessary for Ripple drawable to
    // be projected outside the bounds.
    // We need to get the wrapped drawable to identify if it's a ripple drawable or not.
    final Drawable unwrappedDrawable = ((DefaultComparableDrawable) mDrawable).getWrappedDrawable();
    // We will set the view's background only if it's RippleDrawable
    if (unwrappedDrawable instanceof RippleDrawable) {
      isRippleBackground = true;
    }
}

public boolean isRippleBackground() {
    return isRippleBackground;
}
```

And `ComponentHost` code would be much simpler with this. Just checking `DrawableComponent#isRippleBackground` should be enough to decide whether to set bg or not.

anushreeag: 1. We can move this method inside the DrawableComponent, but we still need to do some operations for getting the RippleDrawable from MountItem. 
2. Since getDrawable() method of DrawableComponent is private, we have do unwrap the drawable from MountItem in ComponentHost and get the Ripple Drawable. Right?
Please suggest.

muraziz: 1. Oh yeah, that's fair.
2. You could change getDrawable to be package-private.

anushreeag: ok, I will try that and update you here. Thanks for your review.

anushreeag: Done. Please review.

pasqualeanatriello:
I was thinking more about this approach and I'm actually worried this might not always work correctly. I'm especially worried about the correctness of avoiding to mounting something that is in the LayoutOuput list. I'm worried that will eventually cause problems. Also, in this diff we are assuming that there will always be a host associated with the RippleDrawable but that is really only true as a side effect of setting a ClickListener.

I think there is an easier approach here. We already set background directly on views without generating a layout output for them if the view is not a component host. We should just use this approach always.
In MountState we check if something is a host view before setting a bunch of attributes:
https://github.com/facebook/litho/blob/19e62e9441d0cfbfb979d300c6609e80241c6afa/litho-core/src/main/java/com/facebook/litho/MountState.java#L1489-L1492
We should move background out of that if.
Here: https://github.com/facebook/litho/blob/9c0e965fd27824f7f535f61712ee06fb40062dea/litho-core/src/main/java/com/facebook/litho/LayoutState.java#L779-L781
We would have to generate a ViewNodeInfo for the HostComponent and make sure that the background gets set there instead of generating a new LayoutOutput.

We would still unfortunately have to unwrap the RippleDrawable but you already have that in this diff.
Does it make sense?

anushreeag:
> I was thinking more about this approach and I'm actually worried this might not always work correctly. I'm especially worried about the correctness of avoiding to mounting something that is in the LayoutOuput list. I'm worried that will eventually cause problems. Also, in this diff we are assuming that there will always be a host associated with the RippleDrawable but that is really only true as a side effect of setting a ClickListener.
> 
> I think there is an easier approach here. We already set background directly on views without generating a layout output for them if the view is not a component host. We should just use this approach always.
> In MountState we check if something is a host view before setting a bunch of attributes:
> https://github.com/facebook/litho/blob/19e62e9441d0cfbfb979d300c6609e80241c6afa/litho-core/src/main/java/com/facebook/litho/MountState.java#L1489-L1492
> 
> 
> We should move background out of that if.
> Here:
> https://github.com/facebook/litho/blob/9c0e965fd27824f7f535f61712ee06fb40062dea/litho-core/src/main/java/com/facebook/litho/LayoutState.java#L779-L781
> 
> 
> We would have to generate a ViewNodeInfo for the HostComponent and make sure that the background gets set there instead of generating a new LayoutOutput.
> We would still unfortunately have to unwrap the RippleDrawable but you already have that in this diff.
> Does it make sense?

1. Yes I clearly understand the code what you have mentioned here. ViewNodeInfo is null for a HostComponent, I observed this while debugging. But HostComponent is created only when we do wrapInView, right? In this case I am getting ComponentHost instead of HostComponent
[LayoutState-CreateLayoutOutput](https://github.com/facebook/litho/blob/master/litho-core/src/main/java/com/facebook/litho/LayoutState.java#L419-L450)
ViewNodeInfo is not null only for MountSpecs.
2. Ripple Drawable is closely associated with clicks only right? So there will be a click listener always associated ?

pasqualeanatriello:
I was suggesting making ViewInfo not null for HostComponents when backgrounds are associated.
It is true that we can expect Ripples to be associated with clicks but that feels like it would be working for a side effect and it would be a bit fragile.

anushreeag:
Fixing the issue in https://github.com/facebook/litho/pull/606 . Will close this PR.

Issue #614, opened 2019-11-25
Kotlin. error: Event fields must be declared as public non-final.


bangonkali:
- [x] I have searched [existing issues](https://github.com/facebook/litho/issues) and this is not a duplicate

## Version
```gradle
    implementation "com.facebook.soloader:soloader:0.6.0"

    implementation "com.facebook.litho:litho-core:$litho_version"
    implementation "com.facebook.litho:litho-widget:$litho_version"
    kapt "com.facebook.litho:litho-processor:$litho_version"
    implementation "com.facebook.litho:litho-fresco:$litho_version"
    implementation "com.facebook.litho:litho-sections-core:$litho_version"
    implementation "com.facebook.litho:litho-sections-widget:$litho_version"
    compileOnly "com.facebook.litho:litho-sections-annotations:$litho_version"
    kapt "com.facebook.litho:litho-sections-processor:$litho_version"
```

## Issues and Steps to Reproduce

**Create a Kotlin Custom Event**
```kotlin
import com.facebook.litho.annotations.Event

@Event
class StorefrontRowMinimizedEvent(
    var storefrontRowId: String = ""
)
```

**Use this event in a Spec**
```kotlin
@LayoutSpec(events = [StorefrontRowMinimizedEvent::class])
object StorefrontItemComponentSpec {
    private const val TAG : String = "StorefrontItem"
```

**You will see error**
```
Event fields must be declared as public non-final.
    private java.lang.String storefrontRowId;
                                         ^
```

**This is how it was compiled in Java**
```java
import java.lang.System;

@com.facebook.litho.annotations.Event()
@kotlin.Metadata(...)
public final class StorefrontRowMinimizedEvent {
    @org.jetbrains.annotations.NotNull()
    private java.lang.String storefrontRowId;
    
    @org.jetbrains.annotations.NotNull()
    public final java.lang.String getStorefrontRowId() {
        return null;
    }
    
    public final void setStorefrontRowId(@org.jetbrains.annotations.NotNull()
    java.lang.String p0) {
    }
    
    public StorefrontRowMinimizedEvent(@org.jetbrains.annotations.NotNull()
    java.lang.String storefrontRowId) {
        super();
    }
    
    public StorefrontRowMinimizedEvent() {
        super();
    }
}
```

**Special emphasis on**

```java
    @org.jetbrains.annotations.NotNull()
    private java.lang.String storefrontRowId;
```

## Expected Behavior

[...]

PR #615, opened 2019-11-26
[EventGenerator] Fix generation from Kotlin class files

pavlospt:
## Summary

This PR resolves the issue that occurs when trying to generate an `EventDeclaration` from an `@Event` annotated Kotlin class. It was brought up on https://github.com/facebook/litho/issues/614 and it makes sure that it can work with `@JvmField` annotated fields & without.

## Changelog

Properly generate Event classes from Kotlin source files. (#614 )

## Test Plan

It was tested manually by adding relevant files on `sample-barebones-kotlin` in order to also ensure future compatibility.


astreet:
Thanks for doing this! I was talking to @colriot and he brought up a good point that since we're special casing Kotlin events already, we can actually just make them immutable data classes. The Java api with mutable fields is just a relic from when we did recycling of event objects.

So something like

```
@Event
data class BoxItemChangedEvent(val newColor: Int, val newStatus: String, val highlightedItemIndex: Int)
```

It would also be awesome if we could make this automatically work for Java classes that are Immutable and take params in the ctor and expose getters. We should be able to treat both the same by just looking for a ctor with params vs. public fields. What do you think?

pavlospt:
@astreet to be honest I am not quite sure if I have understood correctly what you suggest. It would definitely be easier though to just lookup for a ctor with params vs guessing that a Class is a Kotlin generated one!

astreet:
I mean that the code we generate would use the ctor if the ctor exists, otherwise it would enforce that the event type has mutable fields (as it currently does) -- this means it'd work with both data objects and java types that have ctors (we can make it the new recommended pattern!).

I don't believe this should be breaking... was there a case you were thinking of?

pavlospt:
@astreet Cool, that sounds fine then! If it is not enforced with a validation like the existing one on `EventValidation` then it won't be a breaking one. Do you want me to take a look on the ctor approach?

astreet:
Sounds awesome!

pavlospt:
@astreet @colriot My last commit is working around the alternative approach we talked about! Please take a look and let me know :D

litho-processor/src/main/java/com/facebook/litho/specmodels/model/EventValidation.java
```diff
@@ -60,6 +62,13 @@
                 "Event declarations must be annotated with @Event."));
       }
 
+      final boolean isKotlinClass = SpecElementTypeDeterminator.determine(
+          (TypeElement) eventDeclaration.representedObject) == SpecElementType.KOTLIN_CLASS;
+
+      if (isKotlinClass) {
+        continue;
```
astreet: I think we also need to skip this if doing the Java ctor method right? Ideally we'd actually validate that it has the appropriate getters for the fields in the ctor. Even for the kotlin one we might want to validate there's only one ctor since we use the first one in `hasNonEmptyConstructors`

litho-processor/src/main/java/com/facebook/litho/specmodels/generator/EventGenerator.java
```diff
@@ -119,16 +124,103 @@ static TypeSpecDataHolder generateEventDispatcher(EventDeclarationModel eventDec
...
+        if (fieldModel.field.modifiers.contains(Modifier.FINAL)) {
+          continue;
+        }
+
+        final boolean fieldIsPrivate = fieldModel.field.modifiers.contains(Modifier.PRIVATE);
+
+        eventDispatcherMethod
+            .addParameter(fieldModel.field.type, fieldModel.field.name);
+
+        if (isKotlinClass && fieldIsPrivate) {
```
astreet: I would be very tempted to not even support this and instead use the validation to provide a nice error message on how they should use a data class :) I'm very happy if we can have fewer ways of doing the same thing

pavlospt: @astreet I was totally thinking of the validation in combination with this code's removal! Will take a look at it tomorrow and will add the proper validation plus removing this kind of "hack" for Kotlin!

sample-barebones-kotlin/src/main/java/com/facebook/samples/lithoktbarebones/BoxItemChangedEvent.kt
```diff
@@ -0,0 +1,10 @@
+package com.facebook.samples.lithoktbarebones
+
+import com.facebook.litho.annotations.Event
+
+@Event
+data class BoxItemChangedEvent(
```
colriot: Thank you for providing a test case! 
If you feel empowered, you can add test to `litho-processor` module for an `EventGenerator`s methods. But no pressure - we plan to give our tests more love next year anyway, cleaning up the approach :)

litho-processor/src/main/java/com/facebook/litho/specmodels/generator/EventGenerator.java
```diff
@@ -119,16 +122,74 @@ static TypeSpecDataHolder generateEventDispatcher(EventDeclarationModel eventDec
...
+      * Create a list with all the desired arguments to be for the following statement.
+      * These need to be the following:
+      *
+      * 1) Event class name (type declaration)
+      * 2) Event class name (class instantiation)
+      * 3) Event class constructor parameter names.
+      * */
+      final List<Object> ctorStatementArgs = new ArrayList<>();
+      ctorStatementArgs.add(0, eventDeclaration.name);
+      ctorStatementArgs.add(1, eventDeclaration.name);
```
colriot: nit: Can omit index

litho-processor/src/main/java/com/facebook/litho/specmodels/model/EventValidation.java
```diff
@@ -60,14 +64,47 @@
...
+            eventDeclarationGetters.contains(possibleFieldGetter);
+
+        final boolean isValidField = (isPublicField || fieldHasAvailableGetter) && !isStaticField;
+
+        if (!isValidField) {
           validationErrors.add(
               new SpecModelValidationError(
                   fieldModel.representedObject,
-                  "Event fields must be declared as public non-final."));
+                  "Event fields must be declared as public non-final, "
```
colriot: Should we allow `public` fields only for empty constructors? Let's keep this only for old-style's support

litho-processor/src/main/java/com/facebook/litho/specmodels/generator/EventGenerator.java
```diff
@@ -119,16 +122,74 @@ static TypeSpecDataHolder generateEventDispatcher(EventDeclarationModel eventDec
...
+
+      /*
+      * Iterate over the Event fields in order to achieve the following:
+      *
+      * 1) Add the parameters to the method under construction.
+      * 2) Create a parameter name placeholder for JavaPoet.
+      * 3) Keep a list of all parameter names.
+      * */
+      for (FieldModel fieldModel : eventDeclaration.fields) {
+        eventDispatcherMethod
```
colriot: We should skip `static` (`static final`) fields here as well.

litho-processor/src/main/java/com/facebook/litho/specmodels/model/EventValidation.java
```diff
@@ -60,14 +64,47 @@
...
+
+        final String fieldName = fieldSpec.name;
+        final String possibleFieldGetter = "get" +
+            fieldName.substring(0, 1).toUpperCase() +
+            fieldName.substring(1);
+
+        final boolean fieldHasAvailableGetter =
+            eventDeclarationGetters.contains(possibleFieldGetter);
+
+        final boolean isValidField = (isPublicField || fieldHasAvailableGetter) && !isStaticField;
```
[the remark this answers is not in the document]
pavlospt: What shall we generate in that case then?

litho-processor/src/main/java/com/facebook/litho/specmodels/generator/TypeElementUtils.java
```diff
@@ -0,0 +1,50 @@
...
+
+    final boolean hasSingleConstructor = availableCtors.size() == 1;
+
+    if (!hasSingleConstructor) {
+      return false;
+    }
+
+    final ExecutableElement classCtor = (ExecutableElement) availableCtors.get(0);
+
+    return classCtor.getParameters().size() > 1;
```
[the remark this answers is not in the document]
pavlospt: This sounds legit! Will update it!

litho-processor/src/main/java/com/facebook/litho/specmodels/generator/EventGenerator.java
pavlospt: The addition of the index was used because I wanted to showcase that there is a reason for the duplication of that add operation, since it is the same value for the first two arguments in varargs below.

litho-processor/src/main/java/com/facebook/litho/specmodels/model/EventValidation.java
pavlospt: Sure will update it!

litho-processor/src/main/java/com/facebook/litho/specmodels/generator/EventGenerator.java
pavlospt: Sure will update it!

pavlospt:
@colriot I believe I made the right changes according to your feedback. PTAL :)

litho-processor/src/main/java/com/facebook/litho/specmodels/model/EventValidation.java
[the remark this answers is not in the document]
colriot: Nothing, just skipping `static final` fields. I mean, we allowed them to be present in the Event, only that :)

litho-processor/src/main/java/com/facebook/litho/specmodels/generator/EventGenerator.java
colriot: True, totally agree with the intention. Sorry for not getting this right

litho-processor/src/main/java/com/facebook/litho/specmodels/generator/EventGenerator.java
```diff
@@ -119,16 +122,81 @@ static TypeSpecDataHolder generateEventDispatcher(EventDeclarationModel eventDec
...
 
-    for (FieldModel fieldModel : eventDeclaration.fields) {
-      if (fieldModel.field.modifiers.contains(Modifier.FINAL)) {
-        continue;
+        ctorParamNames.add(fieldModel.field.name);
       }
+
+      /*
+      * Create a list with all the desired arguments to be for the following statement.
+      * These need to be the following:
```
colriot: Let's maybe reword a bit?
"
Create a list of all arguments for Event instantiation statement below. Should be the following:
1) Event class name (type declaration)
2) Event class name (constructor call)
3) Event constructor parameters' names
"

pavlospt:
@colriot pushed a fixup according to your feedback :) PTAL!

colriot:
LGTM @pavlospt! Can you, please, fix the test and we are good to merge
Thank you for your work 🤙

pavlospt:
@colriot tests fixed and the latest commits where squashed for simplicity! :)

colriot:
Thanks @pavlospt ! JavaPoet update broke tests internally, I'm figuring out, what's better/faster - to update JavaPoet internally or bring the old version back. Will update you on results

pavlospt:
@colriot reverted the latest changes and opened a new PR for JavaPoet bump https://github.com/facebook/litho/pull/618

colriot:
@pavlospt this broke some internal stuff. I need to fix it and merge them together.

pavlospt:
@colriot ok, let me know if I can help in any way!

colriot:
Hi @pavlospt ! I still have work left to support ABI generating mode for this codegen, but I've already found that we didn't add support for getters for `@FromEvent` annotation. Then fields are not public.

pavlospt:
> Hi @pavlospt ! I still have work left to support ABI generating mode for this codegen, but I've already found that we didn't add support for getters for `@FromEvent` annotation. Then fields are not public.

Not sure I understand what you mean. Is there a problematic use case that you can share with me?

pavlospt:
> `@FromEvent` support for POJO getters is missing.
> 
> UPD (didn't see the previous message at first):
> I.e. when we use `@OnEvent` method with `@FromEvent` params and generate an `onEvent` method inside Component, then we simply use `_event.fieldName` to access the field, but when Event is declared as a POJO or data class, these fields will be private and only getters would be accessible. So, we need to support calling getters instead of accessing field in such cases.
> 
> For example, you can add:
> 
> ```
> @OnEvent(BoxItemChangedEventJava::class)
>  fun onEvent(c: ComponentContext, @FromEvent newColor: Int) {
>  }
> ```
> 
> to `ListItemSpec` and it won't compile.

Great! Will fix it once I return from my travel :)

colriot:
@pavlospt no rush – I'll be on vacation until next year too :) Have great holiday!

pavlospt:
Need to fix some tests due to the addition of tests from: https://github.com/facebook/litho/commit/5fc0145c97913de9bfcfaf6ac1643e4b7a48a136

pavlospt:
@colriot do we have any news for this PR 😄 ?

colriot:
Hi @pavlospt! It's not forgotten, don't worry :) We are planning our Kotlin support effort for this year, I'm making sure that everything will be taken into account, this PR (and underlying internal changes) too of course. Hope to return to ABI support of these changes in first weeks of Feb

PR #628, opened 2020-02-17
Adds shadow offset support to CardShadowDrawable

chggr:
## Summary

This change adds x and y shadow offset support to CardShadowDrawable and
CardShadowSpec. Clients can use these offsets to cast different types of
shadows to their cards. At the moment the offsets are hardcoded in
CardShadowDrawable: the x offset is zero and the y offset is half the
shadow size. This change removes this hard-coding from
CardShadowDrawable and exposes two new properties in CardShadowSpec
(shadowDx and shadowDy) to enable clients to set these offsets.

## Changelog

Added support for x and y shadow offsets to CardShadow.

## Test Plan

I have manually tested this change on a simple UI that comprises a list of cards with each card containing an image on the left and a text on the right. I was able to change the x and y offsets for the card shadow and see the effects on the UI. Using this new functionality I was able to create shadows that come from different light sources above, below, at left and at right of the card.

Please see testing details below and let me know if any further testing is required to accept this PR.


muraziz:
Hey @chggr , thanks for PR! Could you sign our CLA (Contributor License Agreement) so that we can pull this change to our internal repo?

chggr:
Hi @muraziz I have now signed the Facebook CLA, please let me know when this change can be pulled into Litho.

chggr:
Hi @muraziz, please let me know if anything else is required from me to pull this change into Litho. What are the timelines around pulling this change in Litho?

adityasharat:
There seem to be merge conflicts :(

chggr:
I have fixed just fixed the merge conflicts, thanks!

muraziz:
We have internal merge conflicts, resolving them now.

muraziz:
@chggr I would like to add those manual samples in the test plan into our snapshot testing. Could you provide the code of those tests that you used?

muraziz:
@chggr 
Also, I'm adding some javadoc to shadowDx, so that users of this API can understand how this is applied:

```
  /**
   * @param shadowSize
   * @param shadowDx Light source offset applied horizontally, i.e. if shadowDx is 0, the shadows on
   *     the left and right sides are both equal to shadowSize. If shadowDx is negative, the light
   *     source moves to the right side, hence left shadow is increased by shadowDx while right
   *     shadow is decreased by shadowDx. If shadowDx is positive, light source moves to the left
   *     side and left and right shadows are adjusted accordingly.
   * @return Shadow applied to the left side of the element.
   */
  public static int getShadowLeft(float shadowSize, float shadowDx) {
...
```

Does this sound good to you?

chggr:
Thanks @muraziz! The javadoc looks good, similarly for shadowDy. I will try to find the code I used for  manual testing and post here.

muraziz:
@chggr
Also, I'm adding some javadoc to shadowDx/shadowDy, so that users of this API can understand how this is applied:
```
  /**
   * @param shadowSize
   * @param shadowDx Light source offset applied horizontally, i.e. if shadowDx is 0, the shadows on
   *     the left and right sides are both equal to shadowSize. If shadowDx is positive, the light
   *     source moves to the left side, hence left shadow is decreased by shadowDx while right
   *     shadow is increased by shadowDx. If shadowDx is negative, light source moves to the left
   *     side and left and right shadows are adjusted accordingly.
   * @return Shadow applied to the left side of the element.
   */
  public static int getShadowLeft(float shadowSize, float shadowDx) {
  ...


  /**
   * @param shadowSize
   * @param shadowDy Light source offset applied vertically, i.e. if shadowDy is 0, the shadows on
   *     the top and bottom sides are both equal to shadowSize. If shadowDy is positive, the light
   *     source moves to the top side, hence top shadow is decreased by shadowDy while bottom
   *     shadow is increased by shadowDy. If shadowDy is negative, light source moves to the bottom
   *     side and top and bottom shadows are adjusted accordingly.
   * @return Shadow applied to the bottom side of the element.
   */
  public static int getShadowTop(float shadowSize, float shadowDy) {
  ...

```
Does this sound good to you?

muraziz:
@chggr sure, 

one more thing,

while adding customisability is good, I think we should provide good defaults when API users only want to provide shadowPx, but let the framework handle light directions. I looked at material design guidelines (https://material.io/design/environment/light-shadows.html#) and it doesn't explicitly say the light source directions. It seems like the light source is vertically top, horizontally middle? Should we provide defaults like before (i.e. keep left/right same, make top half of shadowPx, make bottom 1.5x of shadowPx), or do you have other suggestions?

chggr:
Regarding the javadoc for shadowDx, I believe we need to substitute "If shadowDx is negative, light source moves to the left" with "If shadowDx is negative, light source moves to the right". The javadoc for shadowDy looks good.

Regarding good defaults, I totally agree. The default is already provided in the PR in CardSpec line 103: final float shadowDy = 0.5f * elevation; This is basically replicating the previous behavior before the PR. There are no defaults defined directly on CardShadowDrawable though.

I also had a look and unfortunately I do not have the manual testing code in a format that I can share here.

muraziz:
> Regarding the javadoc for shadowDx, I believe we need to substitute "If shadowDx is negative, light source moves to the left" with "If shadowDx is negative, light source moves to the right". The javadoc for shadowDy looks good.

Good catch! Will fix.

> Regarding good defaults, I totally agree. The default is already provided in the PR in CardSpec line 103: final float shadowDy = 0.5f * elevation; This is basically replicating the previous behavior before the PR. There are no defaults defined directly on CardShadowDrawable though.

Ah, I see, I just saw other places (internally) where `CardShadowDrawable`'s methods were used directly, maybe we should still apply same adjustment there like in `CardSpec`. Let me do that then.

> I also had a look and unfortunately I do not have the manual testing code in a format that I can share here.

No problem, I can create a simple test with the options you provided.

chggr:
Brilliant, many thanks @muraziz

PR #629, opened 2020-02-18
Replace a commented out @Test with @Test @Ignore

graememorgan:
Omitting @Test looks like a test that someone forgot to annotate and triggers an ErrorProne check; @Ignore is the canonical way to skip a failing test temporarily.


muraziz:
Hey @graememorgan , thanks for PR! Could you sign our CLA (Contributor License Agreement) so that we can pull this change to our internal repo?

graememorgan:
Thanks for the quick reply. Signing your CLA seems to involve a surprising amount of paperwork, so I've asked a colleague to send this tiny patch instead.

PR #638, opened 2020-03-17
Rebind when mountState is unbound.

boduan-coder:
For reference only at current being, will sent out a doc.


astreet:
Is this still relevant?

PR #640, opened 2020-03-19
Reproduce the a11y issue during state update.

boduan-coder:
Talkback doesn't play the updated content via state update. See this sample built in Litho playground activity.

I captured a video here:
https://photos.app.goo.gl/gZ3DteYKofhtLL766


astreet:
Is this still relevant?

PR #680, opened 2020-07-27
Add a componentTag to components to allow arbitrary data to be attached for debugging / testing.

tjaneczko:
## Summary

It is currently not possible in Litho to attach arbitrary information to Components for use in debugging and testing. This information can be incredibly useful for debugging, for instance if it is the data source of props on the component. The changes proposed in this PR would enable access to this underlying data source when debugging through Android Studio, and can potentially allow this information to be exposed to tools like Flipper or Stetho's UI inspector.

## Changelog

Add a componentTag to components to allow arbitrary data to be attached for debugging / testing.

## Test Plan

Added test to CommonPropsTest to verify the underlying InternalNode has its componentTag set. `./gradlew clean test` passes successfully. Additionally verified manually in a project which uses Litho.

### DIFF
--- a/litho-core/src/main/java/com/facebook/litho/CommonPropsHolder.java
+++ b/litho-core/src/main/java/com/facebook/litho/CommonPropsHolder.java
@@ -115,12 +115,12 @@ class CommonPropsHolder implements CommonProps {
 
   @Override
   @Nullable
-  public void getTestTag() {
+  public Object getTestTag() {
     return mTestTag;
   }
 
   @Override
-  public void testTag(Object testTag) {
+  public void testTag(@Nullable Object testTag) {
     mTestTag = testTag;
   }
 
--- a/litho-core/src/main/java/com/facebook/litho/CommonPropsHolder.java
+++ b/litho-core/src/main/java/com/facebook/litho/CommonPropsHolder.java
@@ -54,7 +54,7 @@ class CommonPropsHolder implements CommonProps {
   @Nullable private CopyableLayoutProps mLayoutProps;
   @Nullable private Drawable mBackground;
   @Nullable private String mTestKey;
-  @Nullable private Object mTestTag;
+  @Nullable private Object mComponentTag;
   private boolean mWrapInView;
   @AttrRes private int mDefStyleAttr;
   @StyleRes private int mDefStyleRes;
@@ -115,13 +115,13 @@ class CommonPropsHolder implements CommonProps {
 
   @Override
   @Nullable
-  public Object getTestTag() {
-    return mTestTag;
+  public Object getComponentTag() {
+    return mComponentTag;
   }
 
   @Override
-  public void testTag(@Nullable Object testTag) {
-    mTestTag = testTag;
+  public void componentTag(@Nullable Object componentTag) {
+    mComponentTag = componentTag;
   }
 
   @Override
@@ -652,7 +652,7 @@ class CommonPropsHolder implements CommonProps {
     if ((mPrivateFlags & PFLAG_TEST_KEY_IS_SET) != 0L) {
       node.testKey(mTestKey);
     }
-    node.testTag(mTestTag);
+    node.componentTag(mComponentTag);
 
     if (shouldWrapInView()) {
       node.wrapInView();
--- a/litho-core/src/main/java/com/facebook/litho/CommonPropsHolder.java
+++ b/litho-core/src/main/java/com/facebook/litho/CommonPropsHolder.java
@@ -652,7 +652,6 @@ class CommonPropsHolder implements CommonProps {
     if ((mPrivateFlags & PFLAG_TEST_KEY_IS_SET) != 0L) {
       node.testKey(mTestKey);
     }
-    node.componentTag(mComponentTag);
 
     if (shouldWrapInView()) {
       node.wrapInView();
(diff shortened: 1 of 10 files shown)


litho-core/src/main/java/com/facebook/litho/InternalNode.java
```diff
@@ -250,6 +250,12 @@ InternalNode fullImpressionHandler(
   @Nullable
   String getTestKey();
 
+  /**
+   * Arbitrary data which may be attached to a component for testing.
+   */
+  @Nullable
+  String getTestTag();
```
topwu, line 257: Should't the return value be `Object`?

tjaneczko: Ah yep, I also noticed a few of the other function signatures were incorrect as well. Thanks for catching this! Fixed.

litho-core/src/main/java/com/facebook/litho/DebugComponent.java
```diff
@@ -249,6 +249,17 @@ public String getTestKey() {
...
 
+  /**
+   * @return This component's testTag or null if none is set. Unlike {@link getTestKey}, this
+   *     will return tags for any Component, including Components which are not LayoutNodes.
+   */
+  @Nullable
+  public Object getTestTag() {
+    Component component = mNode.getComponents().get(mComponentIndex);
+    CommonProps props = component.getCommonProps();
+    return props == null ? props.getTestTag();
```
topwu, line 260: Should this be `return props != null ? props.getTestTag() : null;`?

tjaneczko: Yep, sorry, fixed that. I'm not sure how gradle was successfully running the tests with this.

adityasharat:
Would love to see some sample use of this new tag.

I really would not want to call it `testTag` (can be confused with the view `tag`, because we similarly have `key` and `testKey`). Since this is tagging a component, can we instead call it `componentTag`?

tjaneczko:
Sure thing, I had named it `testTag` initially to match the naming of `tag` and `testKey`, but I can definitely see how those could be confusing. I can rename it `componentTag` instead.

Our use case for this is to attach additional information to Components when they are created while Litho's debug mode is enabled. We then can use DebugComponent to traverse the Component hierarchy and retrieve this information at a later time. This information can contain things like data sources which fed into the creation of these Components, being able to see this at runtime is really advantageous for debugging.

Should I add some example usage to this PR? From searching within the Litho repository, it seems like sample usage of DebugComponent itself is pretty sparse, and with this tag being arbitrary data determined by the application it doesn't seem like it can be added to something like [ComponentTreeDumpingHelper](https://github.com/facebook/litho/blob/master/litho-core/src/main/java/com/facebook/litho/ComponentTreeDumpingHelper.java).

I'm also not sure why the CircleCI build is failing, it looks like there's an issue with Buck. Is this something I should be concerned about?

adityasharat:
Thank you sharing the use case. This seems reasonable. Hopefully, this would land successfully at our end and the circle ci issue is a false positive.

litho-core/src/main/java/com/facebook/litho/DefaultInternalNode.java
```diff
@@ -158,6 +158,7 @@
   private @Nullable ArrayList<Component> mComponentsNeedingPreviousRenderData;
   private @Nullable ArrayList<WorkingRangeContainer.Registration> mWorkingRangeRegistrations;
   private @Nullable String mTestKey;
+  private @Nullable Object mComponentTag;
```
adityasharat, line 161: I would recommend changing this into an `Map<Component, Object>` instead. The reasoning is similar to why `mComponents` is a list; because Litho flattens the hierarchy, an InternalNode can represent more than one Component.

litho-core/src/main/java/com/facebook/litho/DefaultInternalNode.java
```diff
@@ -1288,6 +1294,12 @@ public InternalNode testKey(@Nullable String testKey) {
     return this;
   }
 
+  @Override
+  public InternalNode componentTag(@Nullable Object componentTag) {
+    mComponentTag = componentTag;
```
adityasharat, line 1299: If we do make `mComponentTag` a Map then we should lazy initialise it here.

litho-core/src/main/java/com/facebook/litho/CommonPropsHolder.java
```diff
@@ -640,6 +652,8 @@ public void copyInto(ComponentContext c, InternalNode node) {
     if ((mPrivateFlags & PFLAG_TEST_KEY_IS_SET) != 0L) {
       node.testKey(mTestKey);
     }
+    node.componentTag(mComponentTag);
```
adityasharat, line 655: If `ComponentTag` is to be a Map it would be reasonable to add it only if the tag was set (i.e. by using a flag like `PFLAG_TEST_KEY_IS_SET`)

litho-core/src/main/java/com/facebook/litho/DefaultInternalNode.java
tjaneczko: Actually, I don't think we need to copy the component tag into the InternalNode, since access via the DebugComponent is only done through the CommonProps, which also stores the tag. I'll most likely remove this from InternalNode and related classes. If there is some need for accessing the component tag in an Internal Node, it can be added in the future.

adityasharat: I agree, if for the use case you are exploring, you can can access this via `Component.getCommonProps().getComponentTag()` then let us avoid add this to the `InternalNode`.

tjaneczko: Removed component tag from `InternalNode`, should hopefully be a lot simpler now. Thanks for the review btw!

Issue #675, opened 2020-07-06
MountSpec in AppBarLayout have wrong behavior


Tobyyyyyyyyyy:
- [x] I have searched [existing issues](https://github.com/facebook/litho/issues) and this is not a duplicate

## Version
`v0.36.0`

## Issues and Steps to Reproduce
1. Warp a VideoView in MountSpec
2. Use this spec in AppBarLayout
3. Infinite loop call `OnMount` and `OnUnmount`, VideoView can't play.

## Expected Behavior
VideoView play normally

## Link to Code
[Sample](https://github.com/facebook/litho/issues/new)
[My solution](https://github.com/Tobyyyyyyyyyy/LithoInAppBarSample/blob/master/app/src/main/java/com/example/wt/lithoinappbarsample/MyLithoView.kt#L38)

My solution cannot solve the fundamental problem, Hope to get the perfect solution.

muraziz:
Given that this issue was opened a year ago, closing it now as it might not be relevant any more due to changes in Litho code since then. But if this is still relevant please create a new PR.

PR #684, opened 2020-08-08
fixed infinite loop of mounting MountSpec in AppBarLayout

isfaaghyth:
## Summary

- fixed infinite loop of `mounting` MountSpec in `AppBarLayout`
- related issue https://github.com/facebook/litho/issues/675

## Changelog

added `offset` and `shouldRequestLayout()` validation on [LithoView](https://github.com/facebook/litho/blob/master/litho-core/src/main/java/com/facebook/litho/LithoView.java)

## Test Plan

N/A

### DIFF
--- a/litho-core/src/main/java/com/facebook/litho/LithoView.java
+++ b/litho-core/src/main/java/com/facebook/litho/LithoView.java
@@ -888,6 +888,10 @@ public class LithoView extends Host {
 
   @Override
   public void offsetTopAndBottom(int offset) {
+    if (offset == 0) {
+      return;
+    }
+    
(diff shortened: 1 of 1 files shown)


litho-core/src/main/java/com/facebook/litho/LithoView.java
```diff
@@ -888,13 +888,21 @@ public void setHasTransientState(boolean hasTransientState) {
 
   @Override
   public void offsetTopAndBottom(int offset) {
+    if (offset == 0) {
```
nicous, line 891: Why would you want to avoid setting an offset of 0?

PR #685, opened 2020-08-09
Add Opacity Configuration

naufalprakoso:
## Summary
Hi, I'm trying to improve this Open Source by adding an opacity configuration in `drawableColor`. Please help to review my code. If you have any suggestions, feel free to give me comments 😁

## Changelog

- Added opacity configuration in `drawableColor`
- Change `drawableColor` int color parameter to string hex color

## Test Plan

Currently, if we want to add background color in `Decoration`, we must pass integer color in `drawableColor`, like `0xFFFFFFFF`. I have some improvement in that. I changed the integer parameter in `drawableColor` with string hex color. For example, if we want to set the drawable color, we can use `drawableColor("ffffff")`.

And then, if we want to add opacity in drawable color, we can use `drawableColor("ffffff", Opacity.<_NUMBER>)`.

### DIFF
--- a/litho-core-kotlin/src/main/kotlin/com/facebook/litho/Resources.kt
+++ b/litho-core-kotlin/src/main/kotlin/com/facebook/litho/Resources.kt
@@ -16,6 +16,7 @@
 
 package com.facebook.litho
 
+import android.graphics.Color
 import android.graphics.drawable.Drawable
 import androidx.annotation.AttrRes
 import androidx.annotation.ColorInt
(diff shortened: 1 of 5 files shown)


colriot:
Hi @naufalprakoso ! 
Thanks for taking time to work on this PR. Currently Kotlin API is experimental and under development and testing internally. And while API is changing we're not yet ready for external contributions for now. Sorry.

PS: Decoration is already not used in the internal version, for example.

PR #686, opened 2020-08-14
NPE fix for concurrency issues.

gogeta95:
## Summary

We are using this library in a project and get NPEs on a specific configuration because of concurrency issues. This catches the NPE and defaults to false.

## Changelog
- Fix NPEs in a rare scenario

## Test Plan

N/A


pakoito:
If racing is a problem that causes NPE when accessing LayoutState then I believe this change could be better done as 

```
if (mLayoutStateContext == null) {
 LayoutState layoutState = mLayoutStateContext.getLayoutState();
 return layoutState != null && layoutState.isCreateLayoutInProgress();
}
```

and avoid catch'm'all approaches.

PR #695, opened 2020-08-20
Switch circleci to circleci/android:api-28-ndk

muraziz:
Differential Revision: D23240409


colriot:
We moved to GH Actions

PR #703, opened 2020-09-07
Update comments for OnBind, OnUnbind and OnMeasure annotations. 

### LINKED ISSUE #702: Docs for @OnBind and @OnUnbind are not accurate

## Version
`0.37.1`

## Link to Code
https://github.com/facebook/litho/blob/bad7023f2668882a9d01525676d8ef868ce06f09/litho-annotations/src/main/java/com/facebook/litho/annotations/OnUnbind.java#L25-L27

Mentioned in the comments of `OnBind` annotation that it takes Android's `Context` as the first parameter. But when applied,  it throws this error: 

> error: Parameter in position 0 of a method annotated with interface com.facebook.litho.annotations.OnBind should be of type com.facebook.litho.ComponentContext.

And it works with `ComponentContext`. Same goes with `@OnUnbind` annotation. Should the comments be updated or is there something I am missing?

amrro:
## Summary
Update docs to reflect the accurate usage of annotations. 

## Changelog
- Update comments for `OnBind` and `OnUnbind` annotations, Fixes #702.
- Add comment to `OnMeasure` annotation.

### DIFF
--- a/litho-annotations/src/main/java/com/facebook/litho/annotations/OnUnbind.java
+++ b/litho-annotations/src/main/java/com/facebook/litho/annotations/OnUnbind.java
@@ -22,7 +22,7 @@ import java.lang.annotation.RetentionPolicy;
 /**
  * A method annotation used in classes that are annotated with {@link MountSpec}.
  *
- * <p>Methods annotated with {@link OnBind} take a {@code ComponentContext} as the first parameter, the
+ * <p>Methods annotated with {@link OnUnbind} take a {@code ComponentContext} as the first parameter, the
  * Object that the {@link MountSpec} mounts as the second parameter, followed by any number of {@link
  * Prop}s.
  *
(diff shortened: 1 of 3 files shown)


litho-annotations/src/main/java/com/facebook/litho/annotations/OnUnbind.java
```diff
@@ -22,8 +22,8 @@
 /**
  * A method annotation used in classes that are annotated with {@link MountSpec}.
  *
- * <p>Methods annotated with {@link OnUnbind} take an Android Context as the first parameter, the
- * Object that the MountSpec mounts as the second parameter, followed by any number of {@link
+ * p>Methods annotated with {@link OnBind} take a {@code ComponentContext} as the first parameter, the
```
colriot, line 25: -> OnUnbind
and fix `<p>` tag, please

colriot (changes requested):
Thank you! Pointed out some minor typos to fix and we're good to go

litho-annotations/src/main/java/com/facebook/litho/annotations/OnUnbind.java
amrro: Fixed.

PR #714, opened 2020-10-06
[Demo only: do not review]Custom visibilityextension

mihaelao:


cable729:
This is linked to from https://fblitho.com/docs/mount-extensions/#example-custom-visibility-event-processing where MountExtensions are labeled as experimental, and that we should not create our own mount extensions. I believe this is out of date and they are no longer experimental. Can a maintainer confirm this?

PR #731, opened 2020-12-15
Reset scroll position of horizontal scroll view on unmount

katgregory:
## Summary

If a horizontal scroll view is recycled, the "new" view will share the scroll position of the "old" view.

## Changelog

This change resets the scroll position of horizontal scroll views on `unmount` to prevent this recycling bug.

## Test Plan

Verified locally that this one-line change fixes the observed bug. Happy to add further testing if it would be useful.

### DIFF
--- a/litho-widget/src/main/java/com/facebook/litho/widget/HorizontalScrollSpec.java
+++ b/litho-widget/src/main/java/com/facebook/litho/widget/HorizontalScrollSpec.java
@@ -303,6 +303,7 @@ class HorizontalScrollSpec {
       mComponentHeight = 0;
       mScrollPosition = null;
       mOnScrollChangeListener = null;
+      setScrollX(0);
     }
   }
 


litho-widget/src/main/java/com/facebook/litho/widget/HorizontalScrollSpec.java
```diff
@@ -303,6 +303,7 @@ void unmount() {
       mComponentHeight = 0;
       mScrollPosition = null;
       mOnScrollChangeListener = null;
+      setScrollX(0);
```
adityasharat, line 306: We might want to restore it in `#mount()` as well (see line 286).

adityasharat (changes requested):
Can we also reset it in `HorizontalScrollLithoView#mount()`?

vinc3m1:
Looks like there's already logic for restoring in `onMount` in the `onPreDrawListener` @adityasharat 

https://github.com/facebook/litho/blob/43076885d992ae230cdf9163858b59ea117e9bac/litho-widget/src/main/java/com/facebook/litho/widget/HorizontalScrollSpec.java#L198-L205

katgregory:
@vinc3m1 is correct - I've verified that the `OnPreDrawListener` listener restores scroll position when when we remount.

For `VerticalScroll` views, I believe the `setScrollY` logic in the analogous `OnPreDrawListener` in `LithoScrollView` not only restores scroll position on remount but also clears scroll position when the view is recycled. This is why we observe the scroll-sharing bug for horizontal but not vertical scroll views. 

https://github.com/facebook/litho/blob/43076885d992ae230cdf9163858b59ea117e9bac/litho-widget/src/main/java/com/facebook/litho/widget/LithoScrollView.java#L111

PR #733, opened 2020-12-17
Adds @LooperMode.Mode.LEGACY in Litho tests that fail in PAUSED mode

chggr:
## Summary

Some Litho tests fail in the new Roboelectric PAUSED LooperMode and
therefore need to explicitly use the LEGACY mode.

## Changelog

Not a user visible change, only affecting tests.

## Test Plan

This change only affects tests.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -52,7 +52,6 @@ subprojects {
             if (offlineDepsDir != null) {
                 systemProperty 'robolectric.dependency.dir', "${rootDir}/$offlineDepsDir"
             }
-            systemProperty 'robolectric.looperMode', "LEGACY"
         }
     }
 }
--- a/litho-it/src/test/java/com/facebook/litho/ComponentTreePropWithReconciliationTest.java
+++ b/litho-it/src/test/java/com/facebook/litho/ComponentTreePropWithReconciliationTest.java
@@ -17,6 +17,7 @@
 package com.facebook.litho;
 
 import static androidx.test.core.app.ApplicationProvider.getApplicationContext;
+import static org.robolectric.annotation.LooperMode.Mode.LEGACY;
 
 import android.view.View;
 import com.facebook.litho.testing.logging.TestComponentsReporter;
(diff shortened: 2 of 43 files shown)


adityasharat:
Hey @chggr Is there a way to reproduce this issue?

chggr:
Thank you @adityasharat for reviewing this PR! 

My understanding is that from Robolectric 4.4 onwards the default LooperMode is PAUSED:
https://github.com/robolectric/robolectric/blob/15950dfa93e6272624d151722186a32f3275b9b8/annotations/src/main/java/org/robolectric/annotation/LooperMode.java#L13

There are some tests currently in Litho that do not work in LooperMode.PAUSED and they require LooperMode.LEGACY. I believe within Litho the LooperMode is globally set to LEGACY in build.gradle
https://github.com/facebook/litho/blob/0f89b65d508dfa2b951d911b0601b74a3c8d57c7/build.gradle#L55

Unfortunately we are not building Litho using gradle, so these tests are failing for us. This PR is about specifying LooperMode.LEGACY explicitly in the tests that are breaking and need it. We could then remove the global setting in a follow up PR so that the new LooperMode.PAUSED is used everywhere expect in the tests that are actually breaking.

colriot:
@chggr Can you add system prop update into this PR as well, please? So that the switch would be atomic

chggr:
@colriot Thanks for the review! 

I have made the change to remove the robolectric.looperMode property from build.gradle, so that all tests run by default with the new PAUSED LooperMode (which is also the default in Robolectric). Tests that fail with PAUSED can be specifically moved to LEGACY. Please note though that my initial commit only covered tests that were failing up to Litho 0.39 (this is the version we have synced to internally). So there might be further tests that have been added since then and need LEGACY.

colriot:
@chggr oh, that's an important clarification! Let us check how would `master` look with that change

chggr:
@colriot Happy new year!

I have added a couple of extra commits after our latest Litho sync, to explicitly use LooperMode.LEGACY for tests that need it. Please let me know how `master` looks with these changes. There will likely still be some tests that are failing, since we do not sync to Litho head. If you could give me a list of these tests, I can go ahead and add LooperMode.LEGACY to them too.

PR #734, opened 2020-12-17
Avoid mocking Android Views in DynamicPropsTest

chggr:
## Summary
The testDynamicElevationApplied() test in DynamicPropsTest is currently
using Mockito to mock an Android View. This change removes mocking and
uses a fake LithoView extension instead to retrieve the set elevation.

## Changelog

Remove the need to mock an Android View in DynamicPropsTest

## Test Plan

This change only affects tests and is not user visible

### DIFF
--- a/litho-it/src/test/java/com/facebook/litho/DynamicPropsTest.java
+++ b/litho-it/src/test/java/com/facebook/litho/DynamicPropsTest.java
@@ -19,8 +19,6 @@ package com.facebook.litho;
 import static androidx.test.core.app.ApplicationProvider.getApplicationContext;
 import static com.facebook.litho.testing.helper.ComponentTestHelper.mountComponent;
 import static org.assertj.core.api.Java6Assertions.assertThat;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.verify;
 
 import android.annotation.TargetApi;
 import android.graphics.Color;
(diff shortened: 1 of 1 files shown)


litho-it/src/test/java/com/facebook/litho/DynamicPropsTest.java
```diff
@@ -228,21 +221,40 @@ public Component build() {
           protected void setComponent(Component component) {}
         };
 
+    // We are using an old version of Robolectric which has a very limited
```
colriot, line 224: We've recently migrated all our tests to the 4.x version, so I wonder if this is still true at all. Can you check if a plain LithoView would be enough, please?

colriot (changes requested):
@chggr one thing that I'd want to make sure before accepting this. We actually might be able to remove this workaround completely 🤔

litho-it/src/test/java/com/facebook/litho/DynamicPropsTest.java
chggr: That is a great point! I have just used a plain LithoView and it worked beautifully. Thanks for the suggestion

colriot: Thank you for fixing this, Christos !

colriot:
Merging

PR #735, opened 2020-12-17
Add support for getting testKeys set on container Components

chggr:
## Summary

Currently, if a test key is set on a Component which resolves into a
LayoutNode, the corresponding InspectableComponent and DebugComponent
fails to return its testKey.

## Changelog

TestKeys can now be returned from Components that resolve into LayoutNodes.

## Test Plan

Verified locally that the test key is properly returned on any Component.

### DIFF
--- a/litho-core/src/main/java/com/facebook/litho/DebugComponent.java
+++ b/litho-core/src/main/java/com/facebook/litho/DebugComponent.java
@@ -261,6 +261,22 @@ public final class DebugComponent {
   }
 
   /**
+   * Returns this component's testKey or null if none is set.
+   *
+   * <p>Unlike {@link #getTestKey()}, this function can return a test key set on any Component,
+   * including container Components which resolve into LayoutNodes.
+   *
+   * <p>Unlike {@link #getTestKey()}, this function can also return test keys set on individual
+   * Components even when they are all resolved into a single InternalNode.
+   */
+  @Nullable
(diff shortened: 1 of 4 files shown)


litho-core/src/main/java/com/facebook/litho/DebugComponent.java
```diff
@@ -260,6 +260,22 @@ public String getTestKey() {
...
 
+  /**
+   * Returns this component's testKey or null if none is set.
+   *
+   * <p>Unlike {@link #getTestKey()}, this function can return a test key set on any Component,
+   * including container Components which resolve into LayoutNodes.
+   *
+   * <p>Unlike {@link #getTestKey()}, this function can also return test keys set on individual
+   * Components even when they are all resolved into a single InternalNode.
+   */
```
colriot, line 271: Can you please add a similar comment to `getTestKey()` method, so that users of it would know the quirks too?

colriot (approved):
One nit below

litho-core/src/main/java/com/facebook/litho/DebugComponent.java
chggr: Sure, sounds good. Done

colriot: Thank you!

PR #740, opened 2020-12-21
Ignore duplicate transition keys in the same layout

chggr:
## Summary

The existing implementation is throwing an error if there are duplicate,
non autogenerated transition keys in the same layout. This change
exposes a new configuration parameter in ComponentsConfiguration to
ignore duplicate transition keys in layout and sets it to true by
default. It uses this parameter within LayoutState to not throw errors
if a duplicate transition key is detected, but ignore it instead.

## Changelog

Ignores duplicate transition keys in layout by default, rather than throwing error.

## Test Plan

Verified locally that this change is working as intended.


litho-core/src/main/java/com/facebook/litho/config/ComponentsConfiguration.java
```diff
@@ -145,6 +145,9 @@
    */
   public static boolean enableShouldCreateLayoutWithNewSizeSpec = true;
 
+  /** Ignore duplicate transition keys in the same layout, rather than throwing an exception. */
+  public static boolean ignoreDuplicateTransitionKeysInLayout = true;
```
yschimke, line 149: Do we expect some clients to configure this to false? Should this just be the new logic? As a static, it also seems awkward because it's hard to know if it get's changed by another view integrating Litho in the same app.

chggr: I will defer this question to the Litho team. Perhaps the default configuration item should be false to preserve the existing logic and we can change it to true for our use cases.

adityasharat: Yeah, I believe we should probably set it `false`.

adityasharat (changes requested):
Can we set the default to `false`?
Would love to understand why you would want to allow duplicate keys.

litho-core/src/main/java/com/facebook/litho/config/ComponentsConfiguration.java
chggr: Sounds good, I have turned this by default to false to preserve the current logic.

chggr:
I believe currently the transitions code uses @TreeProps to determine if there are duplicate transition keys in the tree hierarchy. This works fine on the first display. If the Component is re-generated though using a previously saved ComponentContext, the latter includes the old TreeProps as well and thus it looks like there are duplicate transition keys. Disabling this flag allows us to avoid throwing errors in this use case.

adityasharat:
> Component is re-generated though using a previously saved ComponentContext ...

This sounds like some custom behaviour; is it? We do not expect the old TreeProps to show up in a new layout, it could be a bug, and I would really appreciate it if you could share a playground to repro this.

chggr:
Hmm, thanks @adithyaakrishna. Let me investigate this a bit further and come back.

PR #742, opened 2020-12-22
Provide the ability to manually set initRangeSize in RecyclerBinder

chggr:
## Summary

At the moment the RecyclerBinder computes the layout synchronously on
the main thread to derive the rangeSize. This change exposes a setter
in the RecyclerBinder.Builder that allows users to set an expected
initial rangeSize and use that to derive the initRange rather than
measure the first component on the main thread.

This is a performance optimisation, that is by default turned off and
only used when the initRangeSize parameter is explicitly set in the
RecyclerBinder builder.

## Changelog

Adds optional performance optimisation in RecyclerBinder to manually
specify the initial rangeSize.

## Test Plan

We have verified that setting an initial range size manually on the 
RecyclerBinder.Builder bypasses the initial layout calculation on the main
thread and avoids the associated performance cost.

### DIFF
--- a/litho-widget/src/main/java/com/facebook/litho/widget/RecyclerBinder.java
+++ b/litho-widget/src/main/java/com/facebook/litho/widget/RecyclerBinder.java
@@ -125,6 +125,7 @@ public class RecyclerBinder
   private final @Nullable LithoViewFactory mLithoViewFactory;
   private final ComponentTreeHolderFactory mComponentTreeHolderFactory;
   private final Handler mMainThreadHandler = new Handler(Looper.getMainLooper());
+  private final int mInitRangeSize;
   private final float mRangeRatio;
   private final AtomicBoolean mIsMeasured = new AtomicBoolean(false);
   private final AtomicBoolean mRequiresRemeasure = new AtomicBoolean(false);
@@ -418,6 +419,7 @@ public class RecyclerBinder
   public static class Builder {
     public static final float DEFAULT_RANGE_RATIO = 2f;
 
+    private int initRangeSize = UNSET;
     private float rangeRatio = DEFAULT_RANGE_RATIO;
     private LayoutInfo layoutInfo;
     private @Nullable LayoutHandlerFactory layoutHandlerFactory;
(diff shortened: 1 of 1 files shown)


litho-widget/src/main/java/com/facebook/litho/widget/RecyclerBinder.java
```diff
@@ -2704,14 +2716,22 @@ void initRange(
                 logger.newPerformanceEvent(mComponentContext, EVENT_INIT_RANGE));
 
     try {
-      final Size size = new Size();
-      holder.computeLayoutSync(mComponentContext, childWidthSpec, childHeightSpec, size);
+      if (mInitRangeSize > 0 && ThreadUtils.isMainThread()) {
```
yschimke, line 2719: Just a guess, since UNSET == -1

```suggestion
      if (mInitRangeSize != UNSET && ThreadUtils.isMainThread()) {
```

chggr: The initial code is correct, we also need mInitRangeSize to not be zero. If it is zero, there will be a division by zero in lines 2722 and 2723

mihaelao:
Hey @chggr , there's a flag on the RecyclerBinder which can be used to manually set the estimated viewport count and it bypasses measuring the first item if it's set: https://github.com/facebook/litho/blob/99eb6d35afacdf9054a15661a8168aabf492bbed/litho-widget/src/main/java/com/facebook/litho/widget/RecyclerBinder.java#L2654-L2656 Can you use that to achieve what you're trying to do?

chggr:
Thanks @mihaelao, that's awesome! I will check whether we can use the estimatedViewPortCount instead and follow up on this PR in the next couple of days

chggr:
Hmm, it seems that the ability to set the estimatedViewPortCount is marked as deprecated. There is a comment stating this is a temporary hack that will soon be removed. Would it be possible to confirm any deprecation plans for this API?

https://github.com/facebook/litho/blob/aa7a2d72a725a3524d11700b586e6fdb43c46b19/litho-widget/src/main/java/com/facebook/litho/widget/RecyclerBinder.java#L645-L657

mihaelao:
It was marked as deprecated to prevent usage in the common case, and it's being used in very specific use cases on critical performance paths where measuring the first item cannot be relied on to estimate the viewport count. There aren't any plans to remove it so if you also need it, I'd recommend updating this PR to remove the deprecated annotation and replace it with a warning that it should be used with caution.

chggr:
That's great news, thanks for confirming! I will switch to using this API internally and verify it is working as expected for us.

chggr:
Thanks again @mihaelao for the insights, I am closing this PR and we will experiment internally with the approach you have kindly suggested. I will create another PR as discussed to un-deprecate the estimated ViewPort count API in the RecyclerBinder

PR #743, opened 2020-12-22
Fix support for disappear transitions

chggr:
## Summary

The current setupTransitions() implementation never looks at any
transitions defined on the previous LayoutState when creating
AnimationBindings. This can be an issue for disappear transitions
because the new LayoutState will no longer contain any disappearing
items.

This change adds logic in the TransitionManager to first create
AnimationBindings for the previous LayoutState, then create
AnimationBindings for the new LayoutState and finally merge the two.

## Changelog

Fixes support for disappear transitions in TransitionManager

## Test Plan

We have manually verified disappear transitions are working as expected
after this change.

### DIFF
--- a/litho-rendercore-transitions/src/main/java/com/facebook/litho/TransitionManager.java
+++ b/litho-rendercore-transitions/src/main/java/com/facebook/litho/TransitionManager.java
@@ -284,10 +284,12 @@ public class TransitionManager {
     List<AnimationBinding> currentAnimationBindings = new ArrayList<>();
     if (currentLayoutState != null) {
       List<Transition> currentTransitions = currentLayoutState.getTransitions();
-      for (Transition transition : currentTransitions) {
-        AnimationBinding binding = createAnimationsForTransition(transition);
-        if (binding != null) {
-          currentAnimationBindings.add(binding);
+      if (currentTransitions != null) {
+        for (Transition transition : currentTransitions) {
+          AnimationBinding binding = createAnimationsForTransition(transition);
+          if (binding != null) {
+            currentAnimationBindings.add(binding);
+          }
         }
       }
     }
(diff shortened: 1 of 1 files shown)


chggr:
Hi @adityasharat I hope you are well! Just wanted to check whether there are any thoughts / feedback on this PR. Thanks!

adityasharat:
Hey @nicous,
Can we collaborate on this PR here?

nicous:
Hey @chggr can you point to a specific case when this would be needed?

chggr:
Hi @nicous, thanks for looking into this PR! This is a transitions issue that we unearthed a while back. If you set a disappearTo value on a Component and then update the ComponentTree so that the Component is gone, the UI shows the Component disappear immediately.

nicous:
Hey @chggr just tried this and it works, maybe there's some other condition. This is the code I used to test this.

```java
@LayoutSpec

  @OnCreateLayout
  static Component onCreateLayout(ComponentContext c, @State boolean noShow) {
    return Column.create(c)
        .backgroundColor(Color.WHITE)
        .clickHandler(PlaygroundComponent.onClick(c))
        .child(
            !noShow
                ? Text.create(c).textSizeSp(20).text("Playground sample").transitionKey("KEY")
                : null)
        .build();
  }

  @OnEvent(ClickEvent.class)
  static void onClick(ComponentContext c, @FromEvent View view) {
    PlaygroundComponent.onUpdateState(c);
  }

  @OnUpdateState
  static void onUpdateState(StateValue<Boolean> noShow) {
    noShow.set(!noShow.get());
  }

  @OnCreateTransition
  static Transition onCreateTransition(ComponentContext c) {
    return Transition.create("KEY").animate(AnimatedProperties.ALPHA).disappearTo(0);
  }
}
```

PR #747, opened 2020-12-23
Adds a new LithoHandler implementation that uses Executors

chggr:
## Summary

This change adds the ExecutorLithoHandler, which is a LithoHandler
implementation that runs all layout computations against a given
Executor. The ExecutorLithoHandler can be used in apps that already
have a well established threading model and would like to use one
of their pre-existing Executors to run all layout against.

## Changelog

Adds a new LithoHandler implementation that uses Executors.

## Test Plan

Unit tests have been added as part of the change.

### DIFF
--- a/litho-core/build.gradle
+++ b/litho-core/build.gradle
@@ -41,7 +41,7 @@ android {
         }
 
         release {
-            buildConfigField 'boolean', internalBuildField, 'false';
+            buildConfigField 'boolean', internalBuildField, 'false'
             buildConfigField 'boolean', useIncrementalMountHelper, 'true'
         }
     }
@@ -70,6 +70,7 @@ dependencies {
 
     // Android Support Library
     compileOnly deps.supportAnnotations
+    implementation deps.guava
     implementation deps.supportCore
     implementation deps.supportCustomView
     implementation deps.supportViewPager
--- a/litho-core/build.gradle
+++ b/litho-core/build.gradle
@@ -70,7 +70,6 @@ dependencies {
 
     // Android Support Library
     compileOnly deps.supportAnnotations
-    implementation deps.guava
     implementation deps.supportCore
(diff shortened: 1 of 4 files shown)


litho-core/build.gradle
```diff
@@ -70,6 +70,7 @@ dependencies {
 
     // Android Support Library
     compileOnly deps.supportAnnotations
+    implementation deps.guava
```
yschimke, line 73: Are there any non-guava alternatives here? Multiset is useful but guava can be a lightning rod dependency. Or is FB Android code ok with Guava?

chggr: There are other places within Litho that are currently using guava, check litho/lib/guava/BUCK

yschimke: I got the impression it was only test code, n.p. if it's already used then.

chggr:
Hi @adityasharat, I hope you are well!

Just wondering whether you have any thoughts or feedback on this PR. Thanks!

adityasharat:
From @astreet 
> I don't think we want to depend on guava in litho-core (unless that's changed?). Even if it has, this change seems like a not great reason to start pulling it in. Maybe this should just go in another module, like a utils one? Or change the impl to not depend on guava and use another datastructure impl.

litho-core/build.gradle
astreet: It's only used in the annotation processor and test code so it doesn't get shipped in any apps using Litho.

chggr: I see, thanks for the context. I will remove usages of MultiSet from the ExecutorLithoHandler, and also the dependency on guava.

chggr:
Many thanks for the feedback! I have now changed the implementation to not depend on Guava at all, as recommended.

PR #751, opened 2021-01-06
Fix litho-core buck files

astreet:
Summary: This was broken in OSS

Differential Revision: D25804645


yschimke:
The contributing guide https://github.com/facebook/litho/blob/master/CONTRIBUTING.md shows options for both Buck and Gradle.  Is Buck preferred, or both should stay working?

PR #754, opened 2021-01-08
Remove unnecessary null checks from DebugComponent

katgregory:
## Summary

Remove two unnecessary null checks from DebugComponent.  In both cases, the second condition of the `if` statement (`Component.isMountViewSpec(component)`) internally performs the same null check ([here](https://github.com/facebook/litho/blob/bc7ebe1a33d9d5e518d975ddb895732f1c25922e/litho-core/src/main/java/com/facebook/litho/Component.java#L798)).

## Changelog

Remove two unnecessary null checks from DebugComponent.

## Test Plan

This quick change only removes redundancy: it does not actually change the logic or UI.


nicous:
Hey!
this looks good! Will import.

PR #756, opened 2021-01-08
Remove unnecessary public modifiers from ClickableSpanListener interface methods

martynawieczorek:
## Summary

The abstract methods in the ClickableSpanListener interface include 'public' modifiers - those are redundant, since all interface abstract methods are implicitly public anyway.
 
## Changelog

Remove the 'public' modifiers from ClickableSpanListener's onClick and onLongClick methods.

## Test Plan

N/A - this change doesn't modify any functionality


nicous:
This looks ok!

PR #761, opened 2021-01-26
Remove LooperMode.LEGACY from tests that do not need it

chggr:
## Summary

Many tests were previously annotated with @LooperMode(LEGACY) because
they were failing when using the new Robolectric PAUSED LooperMode. I have reran
these tests after Litho upgraded to Robolectric 4.4 and some of them no longer
need @LooperMode(LEGACY) and they run successfully without this annotation,
using the new PAUSED LooperMode.

## Changelog

Remove LooperMode.LEGACY from tests that run successfully without it.

## Test Plan

N/A, this change only affects tests.


chggr:
Hi team, just wanted to check whether you had any feedback on this PR. Thanks!

PR #762, opened 2021-01-26
Add @LooperMode LEGACY in tests that fail in PAUSED mode

chggr:
## Summary

The new Robolectric PAUSED LooperMode will become the default in new versions of
Robolectric. We have ran all litho-it tests against the new PAUSED LooperMode and
unfortunately some of them are failing. For these tests we need to explicitly specify the
LooperMode as LEGACY to be future proof, so that they do not start failing when the
default changes to PAUSED in the future.

## Changelog

Add @LooperMode LEGACY in tests that fail in PAUSED LooperMode

## Test Plan

N/A, this change only affects tests


chggr:
Hi team, just wanted to check whether you had any feedback on this PR. Thanks!

PR #807, opened 2021-07-15
Add a reference to the associated EditText view in EditorActionEvent

martynawieczorek:
## Summary

Add a reference in EditorActionEvent to the EditText view associated with the event. This allows the users to access the underlying EditText view when the event is received, similarly to the TextChangedEvent.

## Changelog

Add a reference to the associated EditText view in EditorActionEvent.

## Test Plan

No relevant tests; I confirmed that `dispatchEditorActionEvent` is not called in any other places.

### DIFF
--- a/litho-widget/src/main/java/com/facebook/litho/widget/EditorActionEvent.java
+++ b/litho-widget/src/main/java/com/facebook/litho/widget/EditorActionEvent.java
@@ -17,6 +17,7 @@
 package com.facebook.litho.widget;
 
 import android.view.KeyEvent;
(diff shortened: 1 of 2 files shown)


astreet:
I need to make some changes internally to land this (other references to the event) - since Android's `onEditorAction` takes a TextView and not an EditText I'm going to change the param to a TextView. This also helps me land this since we seem to be only assuming TextView elsewhere in our codebase in the context of dispatching this event.

martynawieczorek:
Thanks Andy, that sounds good to me!

PR #810, opened 2021-07-26
Add a rawInputType parameter to TextInputSpec

martynawieczorek:
## Summary

Add a rawInputType parameter to TextInputSpec. Being able to set rawInputType allows user to achieve a multiline EditText instance with a Done button. The handling of the parameter, i.e. not setting inputType if rawInputType is specified, as well as the param description, is the same as in the old EditTextSpec (litho/litho-widget/src/main/java/com/facebook/litho/widget/EditTextSpec.java). 

## Changelog

Add a rawInputType parameter to TextInputSpec. 

## Test Plan

N/A. Android doesn't expose a getRawInputType method, only setRawInputType. The field has also been added to MaterialTextInputSpec, and I've verified that the methods with modified signatures are not used anywhere else in Litho.


litho-widget/src/main/java/com/facebook/litho/widget/TextInputSpec.java
```diff
@@ -156,6 +156,8 @@
  * @prop editable If set, allows the text to be editable.
  * @prop inputType Type of data being placed in a text field, used to help an input method decide
  *     how to let the user enter text. To add multiline use multiline(true) method.
+ * @prop rawInputType Type of data being placed in a text field, used to help an input method decide
```
colriot, line 159: This doesn't dd a lot of context in terms how it is different from `inputType`. Can you add more specific guidance for users to pick one or another, maybe even references to Android docs.

colriot (changes requested):
This looks good! Can you jus add more clear javadoc comments to this public API, please?

litho-widget/src/main/java/com/facebook/litho/widget/TextInputSpec.java
martynawieczorek: I have added the specific description from the Android docs - hope this is more clear now, but please let me know if there are further changes to make.

PR #812, opened 2021-08-11
Add overScrollMode property to HorizontalScrollSpec and VerticalScrollSpec

martynawieczorek:
## Summary

Add overScrollMode to HorizontalScrollSpec and VerticalScrollSpec. This allows users to set overScrollMode of the underlying scrollViews. The default property value is the Android default, i.e. OVER_SCROLL_IF_CONTENT_SCROLLS. 

## Changelog

Add overScrollMode property to HorizontalScrollSpec and VerticalScrollSpec.

## Test Plan

N/A - this change simply propagates values of overScrollMode to the underlying scrollViews. There are no existing tests for HorizontalScrollSpec or VerticalScrollSpec to modify.

### DIFF
--- a/litho-widget/src/main/java/com/facebook/litho/widget/HorizontalScrollSpec.java
+++ b/litho-widget/src/main/java/com/facebook/litho/widget/HorizontalScrollSpec.java
@@ -16,6 +16,7 @@
 
 package com.facebook.litho.widget;
 
+import static android.view.View.OVER_SCROLL_ALWAYS;
 import static com.facebook.litho.SizeSpec.EXACTLY;
 import static com.facebook.litho.SizeSpec.UNSPECIFIED;
 
@@ -69,6 +70,7 @@ class HorizontalScrollSpec {
   @PropDefault static final boolean scrollbarEnabled = true;
   @PropDefault static final int initialScrollPosition = LAST_SCROLL_POSITION_UNSET;
   @PropDefault static final boolean incrementalMountEnabled = false;
+  @PropDefault static final int overScrollMode = OVER_SCROLL_ALWAYS;
 
   /** Scroll change listener invoked when the scroll position changes. */
   public interface OnScrollChangeListener {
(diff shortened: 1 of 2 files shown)


astreet:
Looks good however I think the default overscroll mode is OVER_SCROLL_IF_CONTENT_SCROLLS (correct me if I'm wrong). So can we change the @PropDefault to that?

martynawieczorek:
I remember seeing a note in the Android documentation about OVER_SCROLL_ALWAYS being the default, but I can't seem to be able to find it now. I did, however, manage to find this [post](https://stackoverflow.com/questions/26872321/default-overscrollmode-value-in-android-view), which implies that the documentation used to incorrectly say that OVER_SCROLL_ALWAYS is the default, while the source code for Android's View set OVER_SCROLL_IF_CONTENT_SCROLLS as the default.

Happy to change to OVER_SCROLL_IF_CONTENT_SCROLLS, but RecyclerCollectionComponentSpec also sets overScrollMode PropDefault as OVER_SCROLL_ALWAYS: https://github.com/facebook/litho/blob/7c587ad46025cda3351a925cf572ddaa0608e2ae/litho-sections-widget/src/main/java/com/facebook/litho/sections/widget/RecyclerCollectionComponentSpec.java#L121

If the Android default is OVER_SCROLL_IF_CONTENT_SCROLLS, shouldn't RecyclerCollectionComponentSpec also use that as PropDefault?

astreet:
It probably should, however I'm just more concerned about changing existing behavior at this point - since we weren't setting an overscroll mode before I want to make sure we keep the default one for everyone already using Vertical/HorizontalScrollSpec

martynawieczorek:
Updated both Vertical and HorizontalScrollSpec to use OVER_SCROLL_IF_CONTENT_SCROLLS.

colriot:
@martynawieczorek thank you for adding this!

PR #893, opened 2022-03-10
Drag-and-drop example using Litho

conca:
I'm trying to implement drag-and-drop using DataDiffSection, but it doesn't work properly. In general, I'm able to drag items within the RecyclerView. However, the interaction feels a bit weird.


adityasharat:
@conca can you sign the CLA, that would enable me to pull this PR into our source control.

conca:
Sorry for the delay. I totally missed your message. This was just an example of drag-and-drop (that is actually not working as expected) for Pasquale to check and identify what is wrong. Do you still need the signature?

PR #896, opened 2022-04-17
Fix for some links not tappable in RTL languages.

AlexanderRBurke:
## Summary

The call to `mLayout.getParagraphDirection(line)` sometimes doesn't return the correct direction. It should not be used from a drawable. Instead this PR passes isRtl via the spec file which uses the screen layout from the context to determine RTL.

### DIFF
--- a/litho-widget/src/main/java/com/facebook/litho/widget/TextSpec.java
+++ b/litho-widget/src/main/java/com/facebook/litho/widget/TextSpec.java
@@ -16,6 +16,8 @@
 
 package com.facebook.litho.widget;
 
+import static android.content.res.Configuration.SCREENLAYOUT_LAYOUTDIR_MASK;
+import static android.content.res.Configuration.SCREENLAYOUT_LAYOUTDIR_RTL;
 import static androidx.customview.widget.ExploreByTouchHelper.INVALID_ID;
 import static com.facebook.litho.SizeSpec.AT_MOST;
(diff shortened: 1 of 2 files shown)


adityasharat:
Hey @AlexanderRBurke in order to proceed you would need to sign the CLA.

litho-widget/src/main/java/com/facebook/litho/widget/TextSpec.java
```diff
@@ -906,7 +908,8 @@ public void textOffsetOnTouch(int textOffset) {
         highlightStartOffset,
         highlightEndOffset,
         clickableSpanExpandedOffset,
-        c.getLogTag());
+        c.getLogTag(),
+        isRtl(c.getAndroidContext()));
```
adityasharat, line 912: The layout direction can be explicitly set as well, and should take precedence over this. You can get the resolved layout direction in `@OnBoundsDefined` from the `ComponentLayout` by invoking `layout.getResolvedLayoutDirection()`. The resolved direction can be returned as an `Output` from `@OnBoundsDefined`.

The default value is `YogaDirection.INHERIT`, so here it would be something like:

```
layout == YogaDirection.INHERIT ? isRtl(c.getAndroidContext()) : layout == YogaDirection.RTL
```

adityasharat (changes requested):
The layout direction can be explicitly set as well, and should take precedence over `isRtl()`.

PR #897, opened 2022-04-25
fix LifecycleStep dependency for litho-it module

lixw1021:
## Summary

- Issue: When building the `litho-it` module, there is a dependency issue that it can't find `LifecycleStep`, 

- Reason: The `LifecycleStep` was moved to test package in `litho-testing` from the `litho-it` main package at [commit](https://github.com/facebook/litho/commit/1b459f22ed96057646da95e35643ab7433a47754), but `LifecycleStep` still used in main package in `litho-it` module. 

## Changelog

- Replace `testImplementation project(':litho-testing')` with `implementation project(':litho-testing')` in `litho-it`'s [build.gradle](https://github.com/facebook/litho/blob/master/litho-it/build.gradle)

## Test Plan

- build `litho-it` module

### DIFF
--- a/litho-it/build.gradle
+++ b/litho-it/build.gradle
@@ -96,6 +96,7 @@ dependencies {
     implementation project(':litho-core-kotlin')
     implementation project(':litho-widget')
     implementation project(':litho-sections-core')
+    implementation project(':litho-testing')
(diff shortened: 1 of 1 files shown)


astreet:
Nice catch, thank you!

PR #917, opened 2022-11-14
Fix two race conditions in RecyclerBinder.

jaceksur:
## Summary

If the collection has not completed initial layout by the time onNewWorkingRange() is called, it will short circuit and not be called again.

If a collection item has not completed initial layout by the time we attempt to check if it belongs to a working range, the call noop-s and is not attempted again.

In each case, save the working range and apply it when the layout completes.

## Changelog

Fix two race conditions in RecyclerBinder.

## Test Plan

Want to get feedback on my approach first.

### DIFF
--- a/litho-widget/src/main/java/com/facebook/litho/widget/ComponentTreeHolder.java
+++ b/litho-widget/src/main/java/com/facebook/litho/widget/ComponentTreeHolder.java
@@ -26,6 +26,8 @@ import com.facebook.litho.Component;
 import com.facebook.litho.ComponentContext;
 import com.facebook.litho.ComponentTree;
 import com.facebook.litho.ComponentTree.MeasureListener;
+import com.facebook.litho.ComponentTree.WorkingRangeAndPositionHolder;
+import com.facebook.litho.ComponentTree.WorkingRangeHolder;
 import com.facebook.litho.ErrorEventHandler;
 import com.facebook.litho.LithoLifecycleListener;
 import com.facebook.litho.LithoLifecycleProvider;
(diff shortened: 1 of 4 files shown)


adityasharat:
You'll have to accept the CLA for us to import this PR

jaceksur:
Done!

litho-widget/src/main/java/com/facebook/litho/widget/ComponentTreeHolder.java
```diff
@@ -116,6 +118,9 @@ interface ComponentTreeMeasureListenerFactory {
   @GuardedBy("this")
   private int mLastRequestedHeightSpec = UNINITIALIZED;
 
+  @GuardedBy("this")
+  private WorkingRangeAndPositionHolder mPendingWorkingRangeAndPosition = null;
```
pentiumao, line 122: Can we add @Nullable to this field to make it Nullsafe?

litho-widget/src/main/java/com/facebook/litho/widget/RecyclerBinder.java
```diff
@@ -221,6 +223,8 @@ public void run() {
   private final RunnableHandler mPreallocateMountContentHandler;
   private final boolean mPreallocatePerMountSpec;
 
+  private @Nullable WorkingRangeHolder mPendingWorkingRange;
```
pentiumao, line 226: This private field is never accessed, can you clarify its purpose?  Thanks!

pentiumao (changes requested):

PR #923, opened 2023-01-01
Website - Features section image alt text displays as [object Object].

GaneshJampa:
**Title:** Website - Features section image alt text displays as [object Object]

**Description:**

Screenshot_20230102_071959

I noticed that when I tried to render an image in the project, the alt text was displayed as [object Object] instead of the expected value - title.

I traced the issue back to the code that was setting the alt attribute of the img element. It was passing React Fragment(object) to the alt attribute, rather than a string.

To fix this, I have added another key 'alt' in the features array, with String as it's value and passed in the Props. This resulted in the expected alt text being displayed.
I have also tried rendering alt attribute this way - alt={title.props.children} which worked. But felt that adding a new key-value pair would be more concise and better.

Please review and consider merging this fix.

**Steps to reproduce:**

Run the project in a browser
Observe the image alt text being displayed as [object Object]

**Expected result:**

The image alt text should be displayed correctly.

**Actual result:**

The image alt text is displayed as [object Object].

### DIFF
--- a/website/src/pages/index.js
+++ b/website/src/pages/index.js
@@ -27,6 +27,7 @@ import styles from './styles.module.scss';
 const features = [
   {
     title: <>Declarative</>,
+    alt: 'Declarative',
     imageUrl: 'images/home-code.png',
     description: (
       <>
@@ -41,6 +42,7 @@ const features = [
   },
   {
     title: <>Asynchronous layout</>,
+    alt: 'Asynchronous layout',
     imageUrl: 'images/home-async.png',
     description: (
       <>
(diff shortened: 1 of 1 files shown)


GaneshJampa:
Admins and Moderators, I have also made another change regarding responsiveness of the Video Iframe in the website. I am requesting you to review and merge this PR as soon as possible. So that I can make another PR for that major change and wouldn't get into conflicts.

astreet:
Hey, thanks for the PR!

GaneshJampa:
Amazing! Thanks @astreet.

PR #924, opened 2023-01-04
Website - Make YouTube video Iframe responsive on mobile screens

GaneshJampa:
## Summary

The YouTube video iframe was overflowing on mobile screens.

I have fixed this and tested the changes. Now the video displays correctly.

Screenshot_20230104_230122

### DIFF
--- a/website/src/pages/index.js
+++ b/website/src/pages/index.js
@@ -111,8 +111,7 @@ function VideoContainer() {
           <h2>Check it out in the intro video</h2>
           <div className={styles.ytVideo}>
             <iframe
-              width="560"
-              height="315"
+              className={styles.ytVideoIframe}
               src="https://www.youtube.com/embed/RFI-fuiMRK4"
               title="Explain Like I'm 5: Litho"
               frameBorder="0"
--- a/website/src/pages/styles.module.scss
+++ b/website/src/pages/styles.module.scss
@@ -94,29 +94,10 @@
 
 // --- css for the video container ---
 
-.ytVideo {
-  width: 560px;
-  height: 315px;
-  margin: auto;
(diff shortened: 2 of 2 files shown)


GaneshJampa:
I have verified that the video looks correct on desktop as well. 
I'm confident that the video will look correct on both mobile and desktop screens.
image

astreet:
Great, thanks - I'm having someone with better knowledge of CSS take a look and review internally!

GaneshJampa:
Awesome! Thank you for your review and for having someone with more expertise in CSS take a look at the changes.

astreet:
Feedback from the eng I had review is below, let me know what you think!

```
That looks a bit overkill and a bit strange with the absolute styling and % padding-top. I think just the following does the trick for you:

.ytVideoIframe {
    width: 100%;
    max-width: 560px;
    aspect-ratio: 16/9;
}

And then remove all other styling on the iframe and surrounding div
```

GaneshJampa:
Well! I am blown away by this approach👏. Thanks for the review.

Here, I'll explain my approach. I have followed the old hack of maintaining aspect ratio with padding top for less than 600px dimension(mobile). But for desktop I have set constant values for height and width.

Regarding, 'aspect-ratio' property in this approach, it's kinda new feature (2 - 3 years old). So I am worried abt it's support for older browsers. But it's simple and straightforward as said. Definitely a very big thumbs up from my side (but consider it's support for old browsers).

astreet:
I'd say let's just do this simple one - I have a feeling most of the folks looking at it will have fairly up-to-date browsers! Thanks for explaining your approach!

GaneshJampa:
Yes that's pretty obvious. I'll make the changes soon.

PR #939, opened 2023-04-19
Use UI state to block certain tail fetches

doncung:
Summary:
Working on internal state management to track some basic UI details. This controller already holds onto two of the fields I wanted, but was keeping the data private. This diff creates getters for this data. 

Differential Revision: D44737041


doncung:
I'm trying to expose some state that is already kept in the class. I added number of items, as I had a need for it. The alternative of trying to get this information from the RecyclerView is quite ugly as it needs to go through layout manager. I thought about keeping and exposing the LayoutInfo alternatively.

PR #959, opened 2023-10-02
Added gap support to rows and columns

stevenkang90:
## Summary

YogaLayout 2.0.0 added support for gaps, this PR plugs that into Litho to allow usage in clients

### DIFF
--- a/litho-core/src/main/java/com/facebook/litho/LithoNode.java
+++ b/litho-core/src/main/java/com/facebook/litho/LithoNode.java
@@ -1158,10 +1158,6 @@ public class LithoNode implements Node<LithoRenderContext>, Cloneable {
     mYogaWrap = wrap;
   }
 
-  public void setGap(YogaGutter gap, Integer size) {
-    mGap = gap;
-    mGapLength = size;
-  }
   public void wrapInView() {
     mForceViewWrapping = true;
   }
--- a/litho-core/src/main/java/com/facebook/litho/LithoNode.java
+++ b/litho-core/src/main/java/com/facebook/litho/LithoNode.java
@@ -75,6 +75,7 @@ import com.facebook.yoga.YogaConstants;
 import com.facebook.yoga.YogaDirection;
 import com.facebook.yoga.YogaEdge;
 import com.facebook.yoga.YogaFlexDirection;
+import com.facebook.yoga.YogaGutter;
 import com.facebook.yoga.YogaJustify;
 import com.facebook.yoga.YogaMeasureFunction;
 import com.facebook.yoga.YogaNode;
@@ -183,6 +184,10 @@ public class LithoNode implements Node<LithoRenderContext>, Cloneable {
   private @Nullable boolean[] mNestedIsPaddingPercent;
 
   protected @Nullable YogaMeasureFunction mYogaMeasureFunction;
+
+  protected @Nullable YogaGutter mGap;
+  protected @Nullable Integer mGapLength;
+
 
   private @Nullable CommonProps.DefaultLayoutProps mDebugLayoutProps;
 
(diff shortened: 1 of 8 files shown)


litho-core/src/main/java/com/facebook/litho/LithoNode.java
```diff
@@ -185,6 +186,10 @@ public class LithoNode implements Node<LithoRenderContext>, Cloneable {
 
   protected @Nullable YogaMeasureFunction mYogaMeasureFunction;
 
+  protected @Nullable YogaGutter mGap;
+  protected @Nullable Integer mGapLength;
```
adityasharat, line 190: We are in the process of generalising the implementation of LithoNode (i.e. decouple Litho and Yoga) so we don't recommend adding new Yoga props to LithoNode. Instead they should be implemented using the `LayoutProps` APIs.

New `gap`, and `gapLength` APIs can be added to the [LayoutProps](https://github.com/facebook/litho/blob/master/litho-core/src/main/java/com/facebook/litho/LayoutProps.kt#L25C24-L25C24) interface. Then implement the new APIs in [CommonProps](https://github.com/facebook/litho/blob/master/litho-core/src/main/java/com/facebook/litho/CommonProps.java) for the component to write to, and [YogaPropsWriter](https://github.com/facebook/litho/blob/master/litho-core/src/main/java/com/facebook/litho/YogaLayoutProps.kt#L30) for the YogaNode to be written to.

The API can then be exposed to Row and Column component builders by adding new setters in (ContainerBuilder)[https://github.com/facebook/litho/blob/master/litho-core/src/main/java/com/facebook/litho/Component.java#L2161]. In the implementation it would call the new API on the common props.

adityasharat (changes requested):
Thank you for contributing to Litho. I have left review comments explaining a more appropriate way of adding a new layout prop. Please let me know if you need more context.

litho-core/src/main/java/com/facebook/litho/LithoNode.java
stevenkang90: Thanks so much for the feedback, I will give this a go and reach out if I need any support

stevenkang90:
@adityasharat I pushed the changes but I'm having issues building the project on master and on the branch to run the unit tests.

```
e: file:///Users/stevenkang/Projects/litho/litho-core-kotlin/src/main/kotlin/com/facebook/litho/FlexboxContainer.kt:133:13 Type mismatch: inferred type is MutableList<Component>? but Style? was expected
```

is there some documentation around this?

adityasharat:
That is odd. It fails to compile with the changes in this PR or it fails to compile even on master?

In this context there are only a couple of places where mutable list of components are used:

* [FlexboxContainerScope](https://github.com/facebook/litho/blob/master/litho-core-kotlin/src/main/kotlin/com/facebook/litho/FlexboxContainer.kt#L146)
* The [Row's](https://github.com/facebook/litho/blob/master/litho-core/src/main/java/com/facebook/litho/Row.kt#L34) or Column's constructor.

stevenkang90:
> That is odd. It fails to compile with the changes in this PR or it fails to compile even on master?
> 
> In this context there are only a couple of places where mutable list of components are used:
> 
> * [FlexboxContainerScope](https://github.com/facebook/litho/blob/master/litho-core-kotlin/src/main/kotlin/com/facebook/litho/FlexboxContainer.kt#L146)
> * The [Row's](https://github.com/facebook/litho/blob/master/litho-core/src/main/java/com/facebook/litho/Row.kt#L34) or Column's constructor.

Hi @adityasharat I've had a few issues building the project to be honest.

1. I needed to change branch from tag rather than master for the project to build
2. I had to bump the c++ version in CMakesList.txt
3. buck is a pain to install on M1 macs, I haven't been able to run the unit tests as it relies on java 8 for buck but my android studio gradle plugin has a minimum of Android 11 (any help with this would be appreciated)

I've managed to deploy and test the changes by integrating them into some feature work so I'm confident it works.

stevenkang90:
Hi @adityasharat were these changes ever released, We deployed an internal artifact to get gap support, recently we updated to 0.49.0 and can't see any gap support in that release still.

PR #981, opened 2024-04-15
Disable docs generation for litho-testing

passy:
Disable docs generation for litho-testing

Summary:
It appears to be timing out in the dokka task here: https://github.com/facebook/litho/actions/runs/8635428192/job/23756165085

Test Plan:
CI

Reviewers:

Subscribers:

Tasks:

Tags:


passy:
Doesn't seem to be necessary.

Issue #978, opened 2024-03-12
Unable to import Litho KComponent, ComponentScope and other Litho classes


connorutting:
- [x] I have searched [existing issues](https://github.com/facebook/litho/issues) and this is not a duplicate

## Version
0.49.0

## Issues and Steps to Reproduce

1. Create a new project in android studio
<img width="400" alt="Screenshot 2024-03-12 at 2 15 46 PM" src="https://github.com/facebook/litho/assets/163178088/3725bfe9-710f-4046-9a52-475d2a7b1048">

2. Select these settings
<img width="500" alt="Screenshot 2024-03-12 at 2 16 16 PM" src="https://github.com/facebook/litho/assets/163178088/a43e08af-b175-4349-b2d6-9287389fae7a">

3. Update your dependencies in the gradle file
```
dependencies {
    implementation("androidx.core:core-ktx:1.9.0")
    implementation("androidx.appcompat:appcompat:1.6.1")
    implementation("com.google.android.material:material:1.11.0")
    implementation("androidx.constraintlayout:constraintlayout:2.1.4")
    testImplementation("junit:junit:4.13.2")
    androidTestImplementation("androidx.test.ext:junit:1.1.5")
    androidTestImplementation("androidx.test.espresso:espresso-core:3.5.1")

    // Litho
    implementation ("com.facebook.litho:litho-core:0.49.0")
    implementation ("com.facebook.litho:litho-widget:0.49.0")

    // SoLoader
    implementation ("com.facebook.soloader:soloader:0.10.5")

    // For integration with Fresco
    implementation ("com.facebook.litho:litho-fresco:0.49.0")

    // For testing
    testImplementation ("com.facebook.litho:litho-testing:0.49.0")
}
```
4. Run gradle sync

5. Modify your `MainActivity.kt` file like so
```kt
import android.os.Bundle
import androidx.appcompat.app.AppCompatActivity
import com.facebook.litho.Component
import com.facebook.litho.ComponentScope
import com.facebook.litho.KComponent
import com.facebook.litho.LithoView
import com.facebook.litho.kotlin.widget.Text
import com.facebook.rendercore.dp

class MyActivity : AppCompatActivity() {

  override fun onCreate(savedInstanceState: Bundle?) {
    super.onCreate(savedInstanceState)

    val lithoView = LithoView.create(this /* context */, MyComponent())
    setContentView(lithoView)
  }
}

class MyComponent() : KComponent() {
  override fun ComponentScope.render(): Component {
[...]

PR #982, opened 2024-04-15
[agp] Upgrade AGP, Kotlin, Gradle, maven publish plugin

passy:
[agp] Upgrade AGP, Kotlin, Gradle, maven publish plugin

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/facebook/litho/pull/982).
* __->__ #982
* #986
* #985

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -30,19 +30,22 @@ POM_DEVELOPER_NAME=facebook
 
 GRADLE_VERSIONS_PLUGIN_VERSION=0.15.0
 
-KOTLIN_VERSION=1.9.10
-DOKKA_VERSION=1.9.10
+KOTLIN_VERSION=1.9.22
+DOKKA_VERSION=1.9.20
 DATA_CLASS_GENERATE_VERSION=1.0.1
 
 # Deps for publishing
-MAVEN_PUBLISH_PLUGIN=0.15.1
+MAVEN_PUBLISH_PLUGIN=0.28.0
 
 # Gradle internals
 org.gradle.internal.repository.max.retries=10
 org.gradle.internal.repository.initial.backoff=1250
-org.gradle.jvmargs=-XX:MaxMetaspaceSize=1024m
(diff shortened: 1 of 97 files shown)


gradle.properties
```diff
@@ -30,12 +30,12 @@ POM_DEVELOPER_NAME=facebook
...
 GRADLE_VERSIONS_PLUGIN_VERSION=0.15.0
 
-KOTLIN_VERSION=1.9.10
-DOKKA_VERSION=1.9.10
+KOTLIN_VERSION=1.9.20
+DOKKA_VERSION=1.9.20
 DATA_CLASS_GENERATE_VERSION=1.0.1
 
 # Deps for publishing
-MAVEN_PUBLISH_PLUGIN=0.15.1
```
zielinskimz, line 38: I think there was an issue with some newer version of this plugin so we reverted back to an old one: https://github.com/facebook/litho/commit/ff08c11f6c8869f6a5f8e36d00e7f0d0713aeabe
Mode details about the issue: https://github.com/facebook/litho/issues/978

passy: Hmm, that's interesting. It seems though as if the old version is no longer able to publish anything now. It times out which looks like there are some problems with the TLS config. https://github.com/facebook/litho/actions/runs/8635428192/job/23756165085

zielinskimz: We were trying 0.22 and you're updating it to 0.28 so I think it may be worth checking if publishing with 0.28 isn't losing some of the classes. WDYT?

passy: Sure! I can push a snapshot locally and if you or someone else could help validate if it's complete, that would be fab.

passy:
Just a note: We may need to revert the AndroidManifest.xml changes because I think Buck builds may still depend on the attribute there. I'll focus on getting the Gradle build to work again first.

gradle.properties
zielinskimz: I guess we could try to use something like this maybe: https://github.com/JakeWharton/diffuse - build a local snapshot using old and new version of the maven publish plugin and compare the apks.

PR #1048, opened 2025-04-29
Remove duplicated Robolectric dependency

MGaetan89:
## Summary

The Robolectric dependency is declared twice in the `:litho-core` module.

## Changelog

Remove duplicated Robolectric dependency.


zielinskimz:
Thanks!

PR #1051, opened 2025-05-19
Fix compilation of the `tests` step on GitHub Actions

MGaetan89:
## Summary

This PR fixes the compilation of the `tests` step on GitHub Actions. Some test dependencies were missing in a couple of modules, preventing the compilation from succeeding.
Note that the step is still failing due to some failing tests.

## Changelog

Add missing test dependencies to `:litho-coroutines-kotlin` and `:litho-rendercore-text`.

## Test Plan

Run the GitHub Actions `tests` step.

PR #1052, opened 2025-05-19
Add more missing testing gradle deps

zielinskimz:
Summary: As per title

Differential Revision: D74989380


MGaetan89:
@zielinskimz I have a similar change in #1051

PR #1053, opened 2025-05-20
Fix Gradle warnings

MGaetan89:
## Summary

This PR fixes some warnings in Gradle files, in particular:
- Use `configureEach` instead of `all` to favor lazy access to objects.
- Use `Task.register()` for creating new tasks.
- Use imports instead of fully qualified class names.
- Move the `dependencies` block out of the `android` block.
- Move the TODO for Lint error back where it belongs.

## Changelog

Fix Gradle warnings.

## Test Plan

GitHub Actions workflow run is successful (minus the currently failing tests).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -14,6 +14,12 @@
  * limitations under the License.
  */
 
+import com.android.build.gradle.tasks.JavaDocGenerationTask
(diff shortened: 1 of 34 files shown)


MGaetan89:
I've updated the PR with the latest changes from `master` and the build is ✅ now.
Let me know if something else is needed for this PR.

PR #1059, opened 2025-06-04
Disable Jetifier

MGaetan89:
## Summary

This PR disables Jetifier. This was used to convert old Android Support Libraries to AndroidX. Nowadays, AndroidX has become the standard, and Android Support Libraries are (almost?) no longer used.

## Changelog

Disable Jetifier.

## Test Plan

Successful build on the CI.


MGaetan89:
@zielinskimz this PR should improve build times by disabling Jetifier.

MGaetan89:
> I wonder if it's possible that the project compiles fine but crashes at runtime if some 3rd party dependency still relies on the old support library. Maybe we should use [plnice/can-i-drop-jetifier](https://github.com/plnice/can-i-drop-jetifier?rgh-link-date=2025-06-10T08%3A22%3A23Z) or [dipien/bye-bye-jetifier](https://github.com/dipien/bye-bye-jetifier?rgh-link-date=2025-06-10T08%3A22%3A23Z) to check if the jetifier can be safely dropped?

Actually, Android Studio provides this feature out of the box. If you open the "Build" panel, select "Warnings" > "Jetifier" and run the check, you should get:
Screenshot 2025-06-10 at 10 30 30

Also, the [plnice/can-i-drop-jetifier](https://github.com/plnice/can-i-drop-jetifier) plugin is deprecated in favor of the native `./gradlew checkJetifier` task, which also says that Jetifier is not needed anymore.

PR #1068, opened 2025-07-02
Fix exhaustive switches

NSProgrammer:
Summary: Changelog: [General][Fixed] - Add `default:` case to avoid warnings/errors for targets that compile with `-Wswitch-enum` and `-Wswitch-default` enabled

Reviewed By: aary

Differential Revision: D77051152


MGaetan89:
@NSProgrammer this seems to have been merged in https://github.com/facebook/litho/commit/f65292e82df2c5046de6e69a5f4960463958ae1f

